Issue #43, opened 2026-05-20
MC64 partial-singular warning bypasses host log verbosity (unconditional eprintln)


jkitchin:
## Summary

The MC64 partial-singular warning is emitted via an unconditional `eprintln!`
straight to stderr. A host application has no way to suppress or redirect it,
so it floods logs even when the host has explicitly requested silent output.

## Where

`src/numeric/factorize.rs`, three identical sites:

- `dense_fast_factor_with_workspace` (~line 1224)
- `factorize_multifrontal_supernodal_with_workspace` (~line 1671)
- `factorize_multifrontal_supernodal_parallel` (~line 2544)

```rust
if let crate::scaling::ScalingInfo::PartialSingular { n_unmatched } = &scaling_info {
    eprintln!(
        "warning: MC64 matching left {} of {} variables unmatched; \
         scaling is identity on those rows/columns",
        n_unmatched, n
    );
}
```

## Why it is a problem

The behavior itself is correct: when MC64 cannot find a perfect matching
(structurally singular matrix), identity scaling on the unmatched
rows/columns is the right fallback, and it mirrors SPRAL/MUMPS.

But `PartialSingular` is *routine* for some callers. feral is used as the
linear-solver backend in pounce (an interior-point NLP solver). An IPM
factorizes the unregularized KKT matrix on the first attempt of every
iteration; that matrix is regularly structurally rank-deficient by a small
amount (free variables that appear only linearly, dependent/redundant
constraints). So `PartialSingular { n_unmatched: 1 }` is expected, benign,
and handled downstream by the IPM inertia correction.

Because the warning is an unconditional `eprintln!`:

- It fires once per affected factorization -- potentially every iteration --
  and floods benchmark logs with an alarming-looking `warning:` line for
  behavior that is normal and recovered.
- It bypasses the host's verbosity control entirely. pounce runs benchmark
  problems with its print level set to silent; feral writes to stderr
  regardless, so the host cannot quiet it short of redirecting stderr
  wholesale.

This is a log-hygiene issue, not a correctness bug.

## Suggested fix

Any of:

1. Route the message through the `log` crate (`log::warn!` / `log::debug!`)
   so hosts control it via a logger. `debug!` is arguably the right level
   given how routine `PartialSingular` is.
2. Add a verbosity / quiet flag to `NumericParams` (or the factor entry
   points) so callers can opt out.
3. At minimum, warn once per `Solver` rather than once per factorization.

[...]

PR #83, opened 2026-06-09
Add whole-repo code review findings document

jkitchin:
What: dev/research/repo-review-2026-06-09.md — consolidated findings
from a six-track audit of src/ (dense, sparse/symbolic/ordering,
numeric, lu, scaling/io/capi, bench) and the six ordering crates,
covering silent bugs, logical flaws vs published algorithms,
performance issues, and cross-path inconsistencies.

Why: capture the full finding set (4 high-severity items incl. silent
inertia corruption in legacy factor() ForceAccept path and a dead
NumericParams::fma feature; ~60 medium/low items) with stable IDs,
file:line evidence, severity/confidence ratings, verified-correct
cores, and a triage order, so fixes can be filed and tracked
individually with tests written first.

Evidence: static review only; no code changed in this commit. Each
finding cites file:line and the contradicting doc/sibling path. The
document itself flags that no finding has a reproducing test yet.

https://claude.ai/code/session_01RnUwLih5FFfmastrxbCrNB

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Numerics — the LU update growth monitor tracks element growth, not the largest single multiplier (L5)
+
+Both the dense (`src/lu/dense_update.rs`) and sparse (`src/lu/sparse_update.rs`)
+Forrest–Tomlin/Bartels–Golub update paths recorded `growth` as the largest
+single elimination multiplier ever seen. A chain of updates each with a
+multiplier of, say, ~100 compounds element growth in `U` to ~100ᵏ while the
+monitor sat flat at 100 — so an ill-conditioned basis could accumulate large
+entries in `U` and silently lose accuracy without ever tripping `max_growth` and
+forcing a refactor. The monitor is now the ‖U‖∞ element-growth high-water ratio
+`max|U| over the update history ÷ max|U| at the last factor`, which compounds
+across updates exactly as standard FT/BG implementations require. This makes the
+monitor strictly stricter (it can only trip earlier, never later), so a basis
+that would have drifted now refactors in time. The sparse path keeps the update
+cheap by scanning only the rows that changed this update — provably equal to the
+true high-water, since every changed `U` entry lives in a changed row. Finding
+from PR #83's review (`dev/research/repo-review-2026-06-09.md`).
+
 ### Scalability — dense-row guard in the sparse LU column ordering (L4)
 
 `SparseColMatrix::ata_pattern` built the explicit AᵀA column-intersection graph
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,25 @@ agreement, and adversarial/ill-scaled/ill-conditioned cases. The downstream
 `pounce-simplex` `BasisEngine` integration and reference (UMFPACK/KLU)
 benchmarks are deferred (see `dev/plans/unsymmetric-lu-epic.md`).
 
+### Fixed — default `postorder()` is now linear, not O(n²·log n), on star etrees (S1)
+
+The default elimination-tree `postorder` (in the standard symbolic pipeline)
+re-cloned and re-sorted `children[node]` on **every** DFS stack visit. A node
+with `c` children sits on top of the stack `c+1` times, so it paid
+O(c²·log c); on a star etree — one root with `n-1` children, the shape AMD
+produces for an arrow/bordered-KKT matrix with a dense *trailing* border —
+the whole traversal was O(n²·log n). It now carries each node's sorted child
+list on the stack (a `(node, sorted_children, cursor)` layout matching the
+already-correct `biased_postorder` / `EliminationTree::postorder`), so each
+node is sorted exactly once and the traversal is O(n·log n).
+
+New regression `test_postorder_star_sort_work_is_linear` reproduces the blow-up
+deterministically (no flaky timing) via a `#[cfg(test)]` work counter: on an
+`n = 2000` star the old code materialized ~`n²` child-list elements (3,998,000);
+the fix materializes ~`n` (≤ `4n`). Output is unchanged — the existing
+topological-order, inverse-roundtrip, and Schur-parity tests still pass.
+Finding S1 from `dev/research/repo-review-2026-06-09.md`.
+
 ### Fixed — `with_fma(true)` now actually dispatches the FMA kernels (N1)
 
 `Solver::with_fma(true)` was a silent no-op. It set `NumericParams::fma`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,22 @@ it, so the factor never stores it; this is recorded in `dev/tried-and-rejected.m
 and pinned as a consistency guard. Finding D4 from
 `dev/research/repo-review-2026-06-09.md`.
 
+### Fixed — contribution-block extraction no longer violates `Vec::set_len`'s init contract (D6)
+
+The contribution-block extraction in `factor_frontal_in_place_with_scratch_impl`
+and `factor_frontal_blocked_in_place_with_scratch` (`src/dense/factor.rs`)
+called `contrib.set_len(cdim²)` **before** writing the cells, then materialized
+a `&mut [f64]` over the still-uninitialized tail. Every cell is written before
+read, so results were correct, but calling `set_len` to expose uninitialized
+elements violates its documented safety precondition (the "write before read"
+property is not the property `set_len` requires). Both sites now initialize the
+region through `spare_capacity_mut()` as `MaybeUninit<f64>` and call `set_len`
+only after all `cdim²` elements are initialized — satisfying the contract at the
+same single-write-per-cell cost (issue #56 Lever B preserved). Output is
+byte-identical (the `blocked_ldlt` scalar-vs-blocked equality suite is
+unchanged). No observable behavior change. Guarded by `tests/d6_contrib_uninit.rs`
+under Miri. Finding D6 from `dev/research/repo-review-2026-06-09.md`.
+
 ### Fixed — static-pivot floor now scale-invariant (computed in scaled space) (N2)
 
 The MA57-style static-pivot floor implied by `static_pivot_threshold = Some(t)`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,22 @@ agreement, and adversarial/ill-scaled/ill-conditioned cases. The downstream
 `pounce-simplex` `BasisEngine` integration and reference (UMFPACK/KLU)
 benchmarks are deferred (see `dev/plans/unsymmetric-lu-epic.md`).
 
+### Fixed — legacy dense `factor()` no longer corrupts the column after a force-accepted zero pivot (D1)
+
+In the legacy dense `factor()` path, a strict-zero pivot routed through
+`ZeroPivotAction::ForceAccept` (and the degenerate 2×2 twin) zeroed its own L
+column but ran no rank-1/rank-2 update, then returned a *fabricated* fused
+next-column argmax of `(0.0, k+2)` / `(0.0, k+3)`. The caller stored that, so
+the next iteration saw `gamma0 == 0.0`, took the "zero off-diagonal column"
+fast path, and silently discarded the **real off-diagonals of the following
+column** — corrupting L (a full-magnitude reconstruction error, ~1.5 on the
+regression matrix) and risking wrong inertia, the project's hard contract.
+Both branches now report the genuine off-diagonal max of the (unmodified) next
+column via `column_offdiag_max`. Regression test
+`tests/dense_ldlt.rs::test_d1_force_accept_does_not_corrupt_next_column`
+asserts exact `P·L·D·Lᵀ·Pᵀ = D_eq·A·D_eq` reconstruction. Finding D1 from
+`dev/research/repo-review-2026-06-09.md`.
+
 ### Added — on-disk dense-column regression fixture for issue #80
 
 `cargo run --bin bench` now regenerates two synthetic dense-coupling-column
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed — `SingularBasis { column }` names the original basis column (L9)
+
+When the sparse LU factor path hit a singular column under
+`LuSingularAction::Fail` (or found no unpivoted row under `PerturbToEps`), it
+reported `FeralError::SingularBasis { column: k }` where `k` is the internal
+*factorization position* — original column `qcol[k]` under the
+(AMD-dependent) column order. A caller such as a simplex driver knows the
+original basis columns it supplied, not the internal processing order, so the
+reported index pointed at the wrong column to repair whenever the column order
+was non-identity. The factor path now reports `qcol[k]`, the original basis
+column, and the `SingularBasis` doc was updated to state this contract
+explicitly. With natural ordering `k == qcol[k]`, so this is a no-op there; it
+only changes the reported index for reordered factorizations. Finding from
+`dev/research/repo-review-2026-06-09.md`.
+
 ### Fixed — sparse and dense LU update report the same failure signal (L8)
 
 `SparseLu::update` returned `FeralError::SingularBasis` when the spike support
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed — sparse singular-column perturbation matches the dense path (L13)
+
+Under `LuSingularAction::PerturbToEps`, the sparse factor perturbed the
+*index-first* still-unpivoted row of a singular column, while the dense factor
+perturbs the *threshold-selected* (largest-|w|) row — so the two paths
+regularized the same singular basis differently (different permutation and
+different regularized solve). The sparse path now perturbs the largest-|w|
+unpivoted row (`ipiv`, the row threshold partial pivoting already selected),
+matching the dense reference. As a side benefit this reuses the pivot already
+found in the selection loop, so the O(m) "find first unpivoted row" scan is
+skipped whenever the singular column has any touched unpivoted entry; the scan
+remains only as a fallback for a column that is structurally empty in every
+unpivoted row. Finding from `dev/research/repo-review-2026-06-09.md`.
+
 ### Fixed — misplaced U diagonal surfaces as an error in release builds (L10)
 
 The sparse triangular solves (`usolve` / `ut_solve`) take the first stored
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Added — `feral-kahip` now reports `KahipStats::n_components` (O18)
+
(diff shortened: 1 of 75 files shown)


jkitchin:
# Verification of the fix campaign (b80af2d..33bcb25)

Independent verification of the 85 fix commits — one audit per finding series, re-reading current code (not just diffs), red-checking new tests by reverse-applying fix hunks, and adversarially re-testing the rejected findings. Full report: [`dev/research/repo-review-2026-06-09-verification.md`](https://github.com/jkitchin/feral/blob/claude/elegant-hawking-0nmp2y/dev/research/repo-review-2026-06-09-verification.md) (c502dc3).

**Bottom line: ~60 of 71 dispositions verified clean — genuinely fixed with real red→green tests, or honestly deferred with accurate `tried-and-rejected.md` entries. Not ship-ready yet: one fix introduced a new high-severity silent-wrong-answer bug, one rejection is partially unsound (concrete repro), one fix is incomplete, and one finding was silently dropped.**

## Action required

### REG-1 (high): N7 fix `131a6de` introduces a silent wrong factorization
`permute_csc_values_cached` now skips the cache *build* when `pattern_reused_hint == false` but never *invalidates* a stale cache, and the warm path validates only `(n, nnz, value_map.len())` — not the pattern (`factorize.rs:3543`). Reproduced at HEAD on one `Solver`: factor pattern A twice (call 2 builds A's cache), then pattern B with the same `(n, nnz)` twice — B-call 2 warm-hits A's stale `value_map`, scatters B's values through A's structure, and returns **Success with solve residual 2.1e+2** (vs 5.7e-14 on B-call 1). The same probe passes at parent commit `0ddaa8b`. Second route: `invalidate_symbolic_cache()` + re-factor where AutoRace picks a different perm for the same pattern. Minimal fix: `*cache = None` on the `hint == false` cold path; durable fix: store a pattern fingerprint in `PermuteCache` and validate it on warm hits.

### REG-2 (high, knob-gated): reopen D5 — `perturb_2x2_to_floor` can create an exactly singular 2×2
The rejection's core proof holds at the default `static_pivot_floor = 0`. But its auxiliary claim that the perturbation "only lifts eigenvalues away from 0" is false: τ is added to both diagonals, which moves a BK block's negative eigenvalue *toward* zero. Demonstrated through the public `factor()` API — 3×3 with leading block `[[-0.5, 1.0],[1.0, -0.5]]`, `A[2,0]=0.25`, `A[2,2]=1.0`, `ForceAccept`, `static_pivot_floor = 2.0` → perturbed block `[[1,1],[1,1]]` (det exactly 0) → unguarded `t = 1/(d00·d11−1)` at `factor.rs:4777` → `d_diag = [1.0, 1.0, NaN]`, ±inf in L, inertia (1, 2, 0). The frontal path re-gates after perturbing; legacy `do_2x2_pivot` does not. Fix: post-perturbation det gate; revisit the D13 deferral with it (its unreachability argument leans on D5's disposition).

### REG-3 (medium): D4 fixed the dense solve gate, not the sparse one
`dense/solve.rs` now correctly gates 2×2 D-block solves on the shared `ssids_det_floor_fail` (verified, red-checked). The production multifrontal solve still uses the old naive `det.abs() > ff.zero_tol_2x2` gate at `src/numeric/solve.rs:281` and `:799` — so the accepted-at-factor-but-skipped-at-solve scenario remains live on the default path, and the dense/sparse solve gates are now mutually inconsistent. No commit or tried-and-rejected entry covers this.

### REG-4 (medium): X2 has no disposition
X2 (MTX declared-nnz never validated against actual entry count; duplicates summed by `to_csc` but overwritten by `to_dense`) is the only finding with no fix, no tried-and-rejected entry, and no journal mention. Compounding it, the X10 fix added `huge_nnz_header_does_not_abort` (`mtx.rs:260-274`), which asserts a file declaring 10^17 entries but containing 2 parses successfully — a future X2 fix must revise that test. X2 was triage-priority 3 (harness integrity).

### REG-0 (fixed): CI-breaking clippy violation
D8's test used `a[0 * n + 2]`, tripping `clippy::erasing_op` under `--all-targets -D warnings`. Fixed in 1785ba2. Root cause worth fixing: the fix sessions ran without the pre-commit hook installed (the exact drift mode CLAUDE.md warns about).

## Smaller residuals (queue)
- **X6 residual hole**: `validate` still accepts `col_ptr[0] != 0` — leading entries silently never factored; one-line check.
- **L2 edge**: sparse diagonal preference (`sparse_factor.rs:316`) lacks the dense path's `> ztol` conjunct, making the formerly-dead `±ztol` clamp reachable (silent sub-tolerance perturbation under `Fail`); `utol` never range-validated.
- **N4 visibility**: the pattern-keyed retry latch can suppress a values-dependent MC64 rescue (wrong inertia where pre-fix it recovered) — acknowledged in the commit message only; belongs in `decisions.md` + the field doc.
- **Doc residue** (~10 sites): `factorize.rs:258` plural-"drivers" permute-cache claim; stale unscaled-norm static-floor docs (`solver.rs:~650`, `factorize.rs:~216`); the bench.rs:1427-1431 dense-KKT rationale comment the X3 rejection itself promised to fix; `feral_capi.h` still claims codes "mirror Ipopt's ESymSolverStatus" and never documents `feral_num_neg`'s −1 sentinel; `error.rs:49-53` "original basis column" claim violated by position-reporting solve paths; stale D7 dispatch comment; D6 test-header Miri claim; kahip `graph.rs:31-33` K4 wiring overstatement.
- **capi `FERAL_SCALING`** still silently ignores unknown values (bench warns; the shim doesn't) — half of X5 remains.
- **Untracked facets**: N3 (`pattern_reused_hint`/`small_leaf` in the parallel driver), N5 (parallel workspaces, warm-permute clone), N1 (direct callers of pub `factorize_multifrontal*` drivers still get a silent `fma` no-op).

## Verified clean (highlights)
D1 corruption fix (red-checked: reconstruction error 1.534 → exact), D2 panel/scalar parity, L1 dense singular-basis guards, L2 CSparse-style diagonal preference, L5 element-growth monitor, S1 postorder rewrite (provably output-identical + deterministic sort-work regression test), N1 FMA dispatch with a genuine engagement test, X4 MC64 unmatched-row fallback (bit-identical when fully matched), X7 zero-clone borrow with a counter-backed test, and all 21 O-series items — including confirmation that the 3ef8967 KKT test change was *not* a tolerance loosening (a bug-pinning assertion replaced by strictly stronger ones, justified in the commit body and journal). The D3 and X3 rejections held up under adversarial re-testing.

Suggested order: REG-1 → REG-2 (+D13) → REG-3 → REG-4 + X6 one-liner → N4 decisions.md entry + doc batch (docs-commit exemption applies).

https://claude.ai/code/session_01RnUwLih5FFfmastrxbCrNB

---

jkitchin:
# Verification of the REG fixes (518b25a, 774b4a3, 1779d81, 8066e93)

All four REG fixes plus the X6 `col_ptr[0]` residual are in. Local gates at HEAD `8066e93`: `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, full test suite passing (63 result groups, exit 0). Each fix shipped as its own atomic commit with a what/why/evidence body, a CHANGELOG entry, and a journal entry; every fix started from a RED→GREEN reproducing test with an external oracle.

## REG-1 (stale permute cache) — fixed correctly
`PermuteCache` now stores the build-time `col_ptr`, `row_idx`, and `perm_inv`, and the warm path compares all three exactly — deliberately stored rather than hashed, so a fingerprint collision can never reintroduce the silent wrong answer. One-shot (`hint == false`) calls also clear any existing cache as defense in depth. Both reproduction routes have RED→GREEN tests oracled against the canonical cold path: same-`(n, nnz)` different pattern, and same pattern under a changed permutation (the AutoRace route). The O(n + nnz) compare remains cheaper than the `from_triplets` sort the warm path skips, so the N7 perf win survives.

## REG-2 (singular perturbed 2×2) — fixed correctly
The subtle part checks out: the guard sits *after* the `swap_rows_cols` that brings row r into position k+1, and `do_2x2_pivot` reads exactly that same `(a[k,k], a[k,k+1], a[k+1,k+1])` triple with no further swap — so the driver's pre-check predicts precisely the block the pivot routine will perturb. The perturbed block is re-gated through the shared `ssids_det_floor_fail` (the same guard the frontal/scalar paths already apply), with a 1×1 fallback that mirrors the existing Duff-Reid `growth_fail` fallback. At the default `static_pivot_floor == 0` the path is byte-identical. The regression test is the verifier's exact counterexample with a solver-independent oracle (a valid LDLᵀ factor never contains NaN/inf). One behavior note, correct in direction: under `Fail` + floor > 0, a perturbed-singular block now returns an error instead of emitting NaN.

D13 was revisited as the verifier requested: it is a distinct path — `do_2x2_update`/`update_2x2_block32` already no-op on `det == 0` and their gated callers divert upstream, so REG-2 only adds the missing divert to the one path (`do_2x2_pivot`) that lacked it. D13's deferral stands unchanged; no D13 code change.

## REG-3 (sparse solve gate) — fixed correctly
Both sparse sites (`solve_sparse_core_into`, multi-RHS `dsolve_node`) route through one shared `solve_2x2_dblock` helper gating on `ssids_det_floor_fail`; the accepted-block arithmetic is the prior kernel verbatim, so accepted blocks invert bit-for-bit and only the gate changed. RED→GREEN test with a hand-computed `rhs = D·x_true` oracle (pre-fix off by 16 orders of magnitude), plus a pin test that factor-rejected blocks stay skipped. The dense and sparse solve gates now agree, closing the D4 inconsistency.

## REG-4 / X2 (+ X6 `col_ptr[0]`) — fixed correctly
The finding that had no disposition now has one, with three reproduced sub-fixes in a single commit:
- **Declared nnz validated** — `parse_mtx` now errors when the size line's `nnz` disagrees with the actual data-line count (oracle: the Matrix Market spec defines that field as the count of entries that follow). RED: a `3 3 4` header with two data lines used to parse to `Ok(len == 2)`.
- **`to_dense` sums duplicates** — it accumulates instead of overwriting, so it matches `to_csc`'s `from_triplets` summing and the Matrix Market / COO convention (scipy, MATLAB). RED: `(1,1)` listed as `2.0` then `1.0` gave `to_dense.get(0,0) == 1.0` (last-wins) while `to_csc` summed to `3.0`; both are now `3.0`. The semantics choice is recorded in `dev/decisions.md` (2026-06-11).
- **`col_ptr[0] != 0` rejected (X6 residual)** — `CscMatrix::validate` now requires `col_ptr[0] == 0`, completing the column-pointer contract the X6 monotonicity check began. RED: `col_ptr = [1, 1, 2]` (monotone, endpoints aligned) passed validation while dropping the leading entry.

The X10 `huge_nnz_header_does_not_abort` test was revised (as the report said it must be): the bogus 10^17-nnz file now returns a recoverable count-mismatch `Err` rather than `Ok`, and the property X10 actually protects — *no process abort* — still holds via the byte-length-clamped reservation (the call returns rather than aborting at `with_capacity`). All four sub-behaviors have RED→GREEN proofs.

## Still open from the verification report

- **Residuals queue**: L2's missing `> ztol` conjunct in the sparse diagonal preference (and unvalidated `utol`); the N4 false-suppression tradeoff still undocumented in `decisions.md`; the ~10-site doc batch (incl. `feral_capi.h`'s stale "mirrors Ipopt" claim and missing −1 sentinel doc); capi's silent handling of unknown `FERAL_SCALING` values; untracked N3/N5 facets.
- Housekeeping: the 2026-06-11 session has a journal but no `dev/sessions/` checkpoint yet — mandatory at session end per protocol.

https://claude.ai/code/session_01RnUwLih5FFfmastrxbCrNB

---

jkitchin:
# Residuals queue cleared — campaign fully discharged (7d13232, 87d5688, 995456f, b99abdd, 079aebe)

The "Still open from the verification report" queue from the [previous comment](https://github.com/jkitchin/feral/pull/83) is now closed. Each shipped as its own atomic commit (RED→GREEN reproducing test where code changed, external oracle, what/why/evidence body, CHANGELOG/journal entry), in the verification report's suggested order. Local gates at HEAD `079aebe`: `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, full suite passing (63 result groups, exit 0); benchmark exit-partition buckets all PASS, no regression.

## L2 — `7d13232` `fix(lu): clear singularity floor in sparse diagonal preference; validate LuParams`
The sparse diagonal preference now carries the dense path's `&& w[diag].abs() > ztol` conjunct, so a diagonal can only be *preferred* when it clears the relative singularity floor — the formerly-dead `±ztol` clamp is replaced with the same `on_singular` routing the dense path uses (`Fail` → `SingularBasis`, `PerturbToEps` → signed floor), closing the silent sub-tolerance perturbation. Added `LuParams::validate()` rejecting `pivot_threshold ∉ (0,1]` and `zero_pivot_tol ∉ [0,1)`, called at the top of dense/sparse `factor`/`refactor`. RED: the new test showed `perm()[0]` diverged from the `DenseLu` oracle pre-fix; GREEN post-fix (`sparse.perm()[0] == 1 == dense.perm()[0]`, solve `Ax=[2,3] → x≈[1,2]`).

## capi `FERAL_SCALING` — `87d5688` `fix(capi): warn on unrecognized FERAL_SCALING, matching bench (X5 follow-up)`
Added the pure predicate `scaling_value_is_unrecognized` (defined via the single shared vocabulary parser, so it can't drift from what the shim accepts) and made `feral_new` emit bench's byte-for-byte warning on a non-empty unrecognized value before keeping the default. Warn-not-error is deliberate: the C ABI's only construction failure channel is a null return, and aborting on a typo'd env var would be hostile (bench also warns-and-continues). RED→GREEN test confirms `""` and the six recognized spellings are not flagged while `mc46`/`infnrom`/`garbage` are.

## N4 / N3 / N5 — `995456f` `docs(n4): record MC64-retry latch inertia tradeoff; track deferred N3/N5 facets`
The `mc64_retry_not_adopted` latch's pattern-keyed-vs-values tradeoff now lives in `decisions.md` and the field doc (not just the original commit message): a pattern that becomes MC64-rescuable at iterate k+1 on different values has its retry suppressed. Bounded safe — MC64 cannot change rank, so the structurally-rank-deficient issue-#43 case the latch targets is exactly safe; residual risk is a numerically-only-singular iterate. Disposition if ever seen to break the inertia gate: make the latch values-aware, not remove it. The deferred N3/N5 parallel-driver facets (permute-cache never engaging on the parallel path, ignored `small_leaf`, parallel-workspace churn, warm-permute O(nnz) clone) now have a `decisions.md` tracking entry. Append-only, no behavior change.

## Doc batch (~10 sites) — `b99abdd` `docs: correct ~10 stale doc/comment sites (repo-review item 4)`
- `feral_capi.h`: corrected the "status codes mirror Ipopt's `ESymSolverStatus`" claim (the exact X8 false claim, now matching `capi.rs:13-24` — only codes 0-2 share Ipopt's values; `FERAL_FATAL=3` collides with `SYMSOLVER_CALL_AGAIN`, shim must translate to `SYMSOLVER_FATAL_ERROR`) and documented `feral_num_neg`'s −1 sentinel.
- static-pivot docs (`factorize.rs`, `solver.rs`, `capi.rs`): `t * ||A||_∞` → scaled `t * ||D·A·D||_∞` (stale post-N2).
- `factorize.rs` `pattern_reused_hint`: qualified "the numeric drivers" to sequential+Schur only — the parallel driver always rebuilds via `permute_csc_values` regardless of the flag (the open N3 facet).
- `bench.rs:1427-1431` dense-KKT rationale: the comment the X3 rejection itself promised to fix — rewritten to the accurate, X3-cited rationale (`params_kkt_dense` is used only by the micro-benchmarks; the validation loop deliberately uses `params_kkt_sparse`; `DIVERGE=0` across 50 parity matrices).
- `error.rs`: `SingularBasis` "original basis column" claim scoped to factor time (solve/update paths report the pivot position); `NeedsRefactor` gained the vanishing-bump / singular-replacement case (L8).
- `dense/factor.rs` D7 dispatch comment: named a function that no longer exists (`factor_block32_in_place_with_scratch`) and inverted which entry is in-place — rewritten (`factor_block32` *is* the in-place pooled-scratch entry).
- `tests/d6_contrib_uninit.rs` header: the Miri-oracle claim contradicted the D6 commit's own honesty note (Miri reports NO UB on the unfixed code — `f64` has no validity invariant; the violated property is the `Vec::set_len` library contract, which Miri does not police) — rewritten to a soundness/finiteness guard.
- `feral-kahip/src/graph.rs` `vweight`: scoped the balance-constraint claim to K3 (K4's flow node separator consumes `vweight` but enforces no balance constraint).

## Housekeeping — `079aebe` `docs(session): checkpoint 2026-06-11-01`
The mandatory session checkpoint (which the previous comment flagged as missing) is written, with refreshed `dev/context.md`.

## Bottom line
Every finding in `repo-review-2026-06-09.md` now has a disposition (87 findings, **0 with no disposition**): each has a fix commit, a `tried-and-rejected.md` deferral/rejection, or a `decisions.md` entry. All five REG regressions (REG-0…REG-4) are fixed; the four low residuals and the housekeeping checkpoint are landed. The campaign is fully discharged.

---

jkitchin:
# Final verification — residuals queue cleared (7d13232..079aebe)

All five follow-up commits verified. Gates at HEAD `079aebe`: `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, full test suite green (63 result groups, zero failures).

1. **`7d13232` (L2 follow-up) — fixed correctly.** The sparse diagonal preference now requires `|w[diag]| > ztol` (dense/sparse parity restored), `LuParams` is range-validated at both factor entry points (rejecting the out-of-contract `pivot_threshold = 0`/NaN and `zero_pivot_tol >= 1` inputs that made the clamp reachable), and the formerly silent `±ztol` clamp now routes through `on_singular` — loud `SingularBasis` under `Fail`, accountable perturbation under `PerturbToEps`. Both behaviors have tests.

2. **`87d5688` (X5 follow-up) — fixed correctly.** The shim now warns on a non-empty unrecognized `FERAL_SCALING` and falls back to default, matching bench. The recognition predicate is defined in terms of the single vocabulary function, so the shim and bench can never drift again; unset stays silent (correct).

3. **`995456f` (N4 + facet tracking) — done properly.** The `decisions.md` entry records the pattern-keyed latch tradeoff with a sound boundedness argument — MC64 cannot change rank, so the structurally-rank-deficient case the latch was built for is exactly safe; the residual numerically-singular-then-rescuable edge is named, with a values-aware remediation path specified if it is ever observed to violate the inertia gate. The solver field doc was updated and the deferred N3/N5 facets got a tracking entry.

4. **`b99abdd` (doc batch) — accurate.** Spot-checked the load-bearing sites: `feral_capi.h` now correctly states the codes are *not* a numeric mirror of Ipopt's `ESymSolverStatus` and that the shim must translate `FERAL_FATAL`; `feral_num_neg`'s −1 sentinel is documented in the C header; the bench dense-KKT rationale comment is rewritten to match the code with the X3 evidence (DIVERGE=0 across all 50 parity matrices) cited; plus the `error.rs` scoping fixes, the D7/D6 comment corrections, and the kahip `vweight` fix.

5. **`079aebe` (session checkpoint) — protocol satisfied.** `dev/sessions/2026-06-11-01.md` written with benchmark results recorded; `context.md` refreshed.

Together with the earlier batches (`8066e93` REG-4/X2 + `col_ptr[0]`, and `518b25a`/`774b4a3`/`1779d81` REG-1/2/3):

**Every finding from the original review (`dev/research/repo-review-2026-06-09.md`) and the verification report (`repo-review-2026-06-09-verification.md`) now has a verified disposition** — fixed with red→green tests, or honestly deferred/rejected with recorded evidence in `tried-and-rejected.md`/`decisions.md`. Nothing actionable remains open on this branch from the review's side.

https://claude.ai/code/session_01RnUwLih5FFfmastrxbCrNB

---

Issue #128, opened 2026-07-10
[perf] Allocation-churn bundle: Schur pack buffers per call, ~9 residual per-update allocations in SparseLu, DenseLu update clones 2m²+m, postorder per-node Vecs; plus supernode nrow underestimate skewing the parallel gate


jkitchin:
Found in the 2026-07-10 full audit (refs are `main` @ `660224d`). Bundled: each item is mechanical pooling/bookkeeping with a shared validation pattern (allocation probes + bit-exactness).

## A. Dense Schur pack buffers allocated per call in the O(n³) hot path
`apply_schur_panel_range_packed` allocates `apack` (`src/dense/factor.rs:3632`) and `bpack0`/`bpack1` (`:3649-3650`) fresh per invocation — once per panel flush, and under `intrafront_parallel` once per rayon chunk. `apack` is `rowspan·n_elim·8B` (≈1.5 MB on a 2955-row root). `bpack1` is allocated and zero-filled even for all-1×1 panels that never read it.
**Fix:** pool in `FactorScratch` (serial) / thread-local scratch (parallel); allocate `bpack1` lazily on first `is_2x2_start`. Byte-exact (layout only). Note: this does NOT reopen the rejected B-1a source-pack idea — the shipped issue-#99 packed micro-kernel stays; only buffer lifetime changes.

## B. Residual per-update allocations in `SparseLu::update_sparse`
Despite the pooling story (`src/lu/sparse_factor.rs:184-186`), ~9 heap allocations remain per update (`src/lu/sparse_update.rs`): `touched` (:124), `supp` (:134), `changed` (:154), `saved_uperm_inv.to_vec()` (:171), `stack` (:334), `reach = touched.clone()` (:357), `old_holders.clone()` (:425), `heap: BinaryHeap::new()` (:506), `offdiag` (:686); plus `new_c`/`new_row` buffers dropped-not-pooled on rollback paths. All poolable exactly like `ft_rw`/`saved_scratch`; the eta `ops` Vec legitimately allocates.
**Fix:** extend the pooled-field pattern; keep `mem::take`/restore discipline on every exit path.

## C. `DenseLu::update` clones `u`, `l`, `qcol` (2m²+m) per update plus `spike` and shift scratch (`src/lu/dense_update.rs:60, 78-80, 161`)
[...]

Issue #99, opened 2026-07-01
Dense-front factorization performance: close the residual ~3.5× gap to faer on large conic KKTs


jkitchin:
## Context

After the `OrderingPreprocess::Auto` fix (#91 / PR #92) took qap15's conic-KKT
factor from ~15 s to ~0.77 s (20×), the **residual ~3.5× gap to faer** (~0.77 s
vs ~0.22 s, 10 cores) is *entirely dense-kernel throughput* on ~36 large fronts,
dominated by a single 2955×2955 indefinite root (42% of the sequential factor
loop). feral now stores *fewer* nonzeros than faer (9.25M vs 13.4M), so this is
per-flop MFLOP/s and parallel scaling, not fill.

This issue tracks the remaining dense-front performance work. Full diagnosis and
design: `dev/research/issue-91-parallel-dense-front-2026-06-30.md` and
`dev/research/issue-91-dense-kernel-profile-2026-06-30.md`. Harnesses:
`examples/{profile_qap15,bench_qap15}.rs`; fixture `tests/data/large/qap15_kkt.mtx`
(gitignored, `dev/scripts/regen_qap15_kkt.sh`).

## Measured facts (don't re-derive)

- **Phase split** (sequential factor, `profile_qap15 phase_split`): trailing
  update (schur) **75%** (1426 ms), assembly **10%** (194 ms), panel factor
  **2%** (41 ms), scalar tail 1.5% (28 ms). ⇒ serial fraction ≈ 0.25.
- **Look-ahead (overlap the serial panel factor) is NOT worth building** — the
  panel factor is only 2%. (Refuted premise; measured, not inferred.)
- **Parallel scaling**: tree-level parallelism ≈ useless here (work is in a few
  big fronts); within-front (`intrafront`) parallelism carries everything but
  scales ~2× on 10 cores (schur itself ~4.7×, sub-linear).
- Ruled out: source-panel packing (net slowdown), larger `block_size` (no change,
  capped at `MAX_N_ELIM=128`), finer intrafront chunks (no gain).

## Done in PR #92 (Lever B, partial)

Lowered `INTRAFRONT_MIN_AREA` 256² → 256×128 — the original ramp-down floor left
each large front's trailing-update tail (and whole medium fronts) on the serial
path. **~17% end-to-end on qap15** (interleaved ~887 → ~739 ms, 10 cores),
byte-exact. Going lower (≤ 16384) regresses (fork/join on tiny fronts); 256×128
[...]

PR #150, opened 2026-08-09
perf(parallel): task-per-subtree coarsening + serial fallback for the parallel driver (#148)

### LINKED ISSUE #148: Parallel multifrontal factorization is slower than serial on 3 of 4 test problems (~1.8M scope.spawn allocations)
## Summary

`factorize_multifrontal_supernodal_parallel_with_workspace` scales **negatively** on most of the problems I tested. Three of four run *faster* with `RAYON_NUM_THREADS=1` than with rayon's default pool. On the worst case the parallel path is **3.1x slower** than serial.

The proximate cause looks like task granularity: the recursive `scope.spawn` in `run_parallel_task` issues ~1.8M spawns for a single solve, each boxing a closure. That allocation traffic is produced on one worker and freed on another, which is the pattern glibc's malloc handles worst — so the cost shows up as allocator contention that grows with thread count.

Found while profiling POUNCE (jkitchin/pounce). Not urgent for me — POUNCE is usable as-is — but the serial path being faster seems worth knowing about.

## Reproduction

POUNCE 0.9.0 (feral 0.14.0), release build, x86_64 Linux (Debian trixie container), 14 cores. Best of 2-3 runs each, interleaved.

Wall clock, glibc allocator, varying `RAYON_NUM_THREADS`:

| problem   | 1 thread | 2 | 4 | default (14) | penalty vs serial |
|-----------|---------:|--:|--:|-------------:|------------------:|
| sparseqp  | 5675 ms | 6858 | 8222 | 17854 ms | **3.14x slower** |
| optcontrol| 5772 ms | 5974 | 6116 | 7033 ms | 1.21x slower |
| bratu     | 156 ms | 164 | 172 | 222 ms | 1.42x slower |
| poisson   | 13220 ms | 9485 | 8299 | 9577 ms | 0.72x (only win) |

`poisson` is the only problem where parallelism pays. Monotonic degradation with thread count on `sparseqp` (5675 -> 6858 -> 8222 -> 17854) is the clearest signal.

Problems are from POUNCE's `benchmarks/large_scale` suite (Pyomo-generated `.nl`).

## Allocation profile

heaptrack on `sparseqp`, unmodified binary:

```
calls to allocation functions: 32516125 (757192/s)
temporary memory allocations:   6818532 (158780/s)
peak heap memory consumption:    708.27M
total runtime: 42.94s
```

The dominant stack, and the reason I landed here:

```
1800002 calls with 0B peak consumption from:
    rayon_core::registry::in_worker
      at rayon-core-1.13.0/src/registry.rs
    rayon_core::scope::scope
      at rayon-core-1.13.0/src/scope/mod.rs:282
    feral::numeric::factorize::factorize_multifrontal_supernodal_parallel_with_workspace
      at feral-0.14.0/src/numeric/factorize.rs:3214
    feral::numeric::factorize::factorize_multifrontal_parallel_with_workspace
```

**1.8M allocations with 0 B peak consumption** — allocated and freed immediately, holding nothing. That is spawn bookkeeping, not data.

To be precise about attribution: `rayon::scope` at :3214 is called **once**, wrapping the loop over leaves. heaptrack attributes these to that frame because the allocations happen beneath it. The actual source is the `scope.spawn(move |s| { ... })` at :3352 in `run_parallel_task`, plus its recursive re-spawn — each spawn boxes a closure.

Separately, 61% of allocations under one `Vec::with_capacity` -> `collect()` stack are temporary (2.9M of 4.7M).

## Evidence it is allocator contention, not raw volume

Same binary, same target, allocator swapped to mimalloc as `#[global_allocator]`, `sparseqp`:

| threads | glibc | mimalloc | ratio |
|--------:|------:|---------:|------:|
| 1 | 5582 ms | 5281 ms | 1.05x |
| 2 | 6861 ms | 5355 ms | 1.28x |
| 4 | 8204 ms | 5928 ms | 1.38x |
| 8 | 8902 ms | 6155 ms | 1.44x |
| default (14) | 17823 ms | 6274 ms | **2.84x** |

At one thread the allocators are equivalent (1.05x) — so it is not allocation *volume* per se. The gap opens monotonically with thread count, which is the signature of cross-thread alloc/free: glibc serializes on arena locks, mimalloc's per-thread free lists do not.

Note that mimalloc **also** degrades with threads (5281 -> 6274 ms), just far less. So a faster allocator masks the symptom rather than fixing it — even under mimalloc, serial beats parallel here.

Raw allocation count is not the discriminator either: `poisson` does 18.2M allocations to `sparseqp`'s 32.5M, but sees no mimalloc benefit at all, consistent with it not going through this path as heavily.

## Suggested directions

Ordered by how much I would trust them, and all pure-Rust with no new deps:

1. **Coarsen task granularity.** 1.8M spawns for one solve suggests a spawn per supernode regardless of its work. A cutoff — below N flops or N columns, factor the subtree inline rather than spawning — would likely remove most of them. This is the fix I would expect to matter most, and it helps every allocator and platform rather than just Linux/glibc.
2. **A serial threshold for the whole path.** Given three of four problems are faster serial, falling back to the serial factorization unless the tree offers enough parallel work looks like it would be a straight win today.
3. **Reuse the `collect()` temporaries.** The 2.9M temporary `Vec`s look like per-call buffers that could be workspace-owned and `clear()`ed. `thread_ws` already exists, so there may be a natural home.

I have deliberately not sent a PR — you know this code and the intended task decomposition far better than I do, and (1) in particular is a design call about where the cutoff belongs.

## What I am not claiming

- I have not profiled with `perf` to confirm where the time actually goes; the thread-scaling and allocator-swap evidence is indirect, though consistent.
- Only four problems, one machine, one architecture.
- I have not checked whether `feral` on `main` already differs from the published 0.14.0.

Happy to run more problems, a `perf` profile, or a specific experiment if any of the above would help.

jkitchin:
Fixes #148 (pending validation on the original POUNCE problems — see "What to verify" below).

## Summary

The parallel multifrontal driver spawned **one boxed rayon task per supernode** — ~1.8M spawn allocations per POUNCE `sparseqp` solve, glibc arena contention growing with thread count, and parallel slower than serial on 3 of 4 problems. This PR implements the issue's suggested directions 1 + 2:

1. **`TaskPlan` subtree coarsening** — bottom-up subtree flop estimates (one postorder pass over supernode geometry), task boundaries only at tree roots and at children of nodes with **≥ 2 sibling subtrees each ≥ `FERAL_PAR_TASK_MIN_FLOPS`** (default 1e6). A lone big child continues inline in its parent's task, so chain-shaped trees collapse to one task per root (the naive `subtree ≥ cutoff` rule made 6319 tasks of 6564 supernodes on a banded-QP proxy; the sibling rule makes 1). Each task factors its owned nodes serially in postorder, then trampolines its parent task via task-children pending counters.
2. **Serial fallback** — a task graph with < 2 seeds has no initial parallelism (an effective chain of tasks), so the driver delegates to the sequential path outright, keeping `intrafront_parallel = true` so wide fronts still fork internally.

`FERAL_DEBUG_TASK_PLAN=1` dumps the plan shape (supernodes / tasks / seeds / cutoff) for triaging future parallel-performance reports.

## Evidence — paired A/B (updated methodology)

Measured by **paired alternating A/B**: old and new run back-to-back within one time window so slow host drift cancels, `min_us` per invocation, ratio < 1 means new is faster. (The original unpaired numbers in this PR were re-measured after review — see the correction note below.)

| proxy | pairs favouring new | median ratio | reading |
|---|---|---:|---|
| grid250 (poisson-like, n=62500) | **10 / 10** (p≈0.002) | 0.762 | new **~24% faster** |
| sparseqpL (banded KKT, n=105000) | 9 / 10 (p≈0.02) | 0.894 | new **~11% faster** |
| chainW (wide-block chain, n=36000) | 9 / 12 | 0.961 | new ~4% faster |

Spawn counts: grid250 **11,171 → 51 tasks** (26 seeds); chains → 1 task → sequential path. The issue's signature — parallel losing to serial on the sparse-QP shape — is gone.

A within-binary control isolates the granularity effect itself on the chain shape: coarse (shipped) vs fine-grained per-node tasks, 9/10 pairs favour coarse, median ratio 1.045 (sign test p≈0.02) — **per-node spawning is ~4% slower** there.

## Bit-exactness

Scheduling-only: per-supernode factor calls and extend-add child order are untouched, so the existing value-determinism argument (decisions.md 2026-05-22) applies to any task partition. New `tests/task_plan_parity.rs` pins fine-grained (cutoff=1: 153 tasks/132 seeds), default, and fallback configurations **byte-identical** to the sequential driver; full suite green (84 test binaries) including `parallel_parity` and the golden bit-digests.

## Correction: the "chainW anomaly" is withdrawn

An earlier revision of this PR documented an unexplained chainW result (per-node spawning appearing ~20% faster than sequential) and a possible 5–18% regression. **Both were measurement noise and are withdrawn.** Proof: three `FERAL_PAR_TASK_MIN_FLOPS` values that produce *identical* task plans (`n_tasks=1, seeds=1` — same code path, byte-identical work) measured 139.6 / 259.1 / 155.5 ms, a 1.9× spread; eight runs of one fixed config spanned 23–31%. Paired re-measurement reverses the sign, as tabulated above. The `AtomicLockStats` reading previously cited (`factor_body` 912 vs 1276 ms) was likewise unpaired and should not be cited.

This produced a methodology change now recorded in `decisions.md`: perf claims in a shared container require **paired alternating A/B, ≥10 pairs, sign test**, `min_us` per sample — the prior "3-run median for sub-5% claims" rule is necessary but not sufficient, since three consecutive medians can all land inside one drift excursion. Full analysis: `dev/research/issue-148-parallel-task-granularity.md`.

## What to verify before/after merging

The definitive acceptance run is the original benchmark set (sparseqp / optcontrol / bratu / poisson, 14 threads, glibc): sparseqp should drop from ~17.9 s toward its ~5.7 s serial time, poisson should keep its parallel win, and heaptrack should show the 1.8M-allocation `scope::spawn` stack gone.

Deferred, tracked in the research note: issue suggestion 3 (the 2.9M `collect()` temporaries — re-profile after this lands, likely partially subsumed) and the #128 `nrow` underestimate, which is now load-bearing for *where* task boundaries fall rather than a yes/no gate and deserves a sensitivity check.

https://claude.ai/code/session_01F87aJtwLT34CTG8KLj5fqm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,26 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed — the profiler no longer reports small fronts as costless (issue #148)
+
+- **Breaking (diagnostic API).** `SupernodeTiming` now records **nanoseconds**:
+  `us` → `ns`, and `assembly_us`/`densefactor_us`/`panelfactor_us`/`schur_us`/
+  `scalartail_us` → the matching `*_ns`. `BucketStats::sum_us`/`avg_us` →
+  `sum_ns`/`avg_ns`, and `ProfileReport::loop_us` is now `loop_ns`.
+  Microsecond accessors (`SupernodeTiming::us()`, `BucketStats::sum_us()`,
+  `ProfileReport::loop_us()`, …) are provided for existing consumers — the
+  migration is `.us` → `.us()`.
+- Why: the per-supernode timer used `Duration::as_micros()`, and the phase
+  deltas were divided by 1000, so every front costing under a microsecond
+  recorded **zero**. On a 250×250 grid Laplacian, **5651 of 11171 supernodes
+  (50.6%) recorded 0 µs**; the aggregate under-reported the inner loop by
+  6.7 ms of 152 ms (4.4%), and by 12.0% on a tridiagonal n=100000 chain. The
+  instrument reported exactly the population it was pointed at — the
+  small-front bucket — as free, which blocked any measurement of amalgamation
+  or per-front fixed-cost work. `loop_ns` now sums nanoseconds, so that
+  population is visible: the `<=8` bucket on grid250 shows 8156 fronts at
+  1.47 µs mean, 7.9% of loop time.
+
 ### Performance — parallel factorization no longer loses to serial (issue #148)
 
 - The parallel multifrontal driver now spawns one rayon task per *subtree*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Performance — parallel factorization no longer loses to serial (issue #148)
+
+- The parallel multifrontal driver now spawns one rayon task per *subtree*
+  (>= 1e6 estimated flops, `FERAL_PAR_TASK_MIN_FLOPS` override) instead of one
+  boxed task per supernode (~1.8M allocations per solve on chain-structured
+  problems, glibc arena contention growing with thread count). Chain-shaped
+  trees collapse to a single task and take the sequential driver outright.
+  Measured (4-core x86_64): the sparse-QP proxy that lost 15-25% to serial at
+  4 threads now matches serial; the grid/poisson-class proxy improves a
+  further 9-21% at 4 threads. Scheduling-only — factors are byte-identical
+  (new `tests/task_plan_parity.rs` gate). `FERAL_DEBUG_TASK_PLAN=1` dumps the
+  task-graph shape for diagnosing parallel-performance reports.
+
 ### Performance — x86 SIMD kernels actually reach AVX2 (session 2026-08-09)
 
 - The x86_64 branches of the pulp dispatch helpers now route through
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ All notable changes to FERAL will be documented in this file.
   `sum_ns`/`avg_ns`, and `ProfileReport::loop_us` is now `loop_ns`.
   Microsecond accessors (`SupernodeTiming::us()`, `BucketStats::sum_us()`,
   `ProfileReport::loop_us()`, …) are provided for existing consumers — the
-  migration is `.us` → `.us()`.
+  migration is `.us` → `.us()`. **The Python API is unchanged**: `sum_us`,
+  `avg_us` and `loop_us` keep their names, types and values, and new
+  `sum_ns`/`avg_ns`/`loop_ns` fields expose the full precision.
 - Why: the per-supernode timer used `Duration::as_micros()`, and the phase
   deltas were divided by 1000, so every front costing under a microsecond
   recorded **zero**. On a 250×250 grid Laplacian, **5651 of 11171 supernodes
@@ -36,8 +38,18 @@ All notable changes to FERAL will be documented in this file.
   further 9-21% at 4 threads. Scheduling-only — factors are byte-identical
   (new `tests/task_plan_parity.rs` gate). `FERAL_DEBUG_TASK_PLAN=1` dumps the
   task-graph shape for diagnosing parallel-performance reports.
+- New `PAR_MIN_SEEDS` (default 2, `FERAL_PAR_MIN_SEEDS` override): how many
+  independent seed tasks the tree must offer before the task graph is used at
+  all. Calibrated on six real KKT matrices — the default is correct, and
+  forcing the task graph on anyway (`=1`) costs 3–9% on chain-shaped trees,
+  while raising it to 4+ costs 28% on a wide tree. Byte-identical at every
+  setting.
+- The sequential-fallback path no longer builds a task plan it discards: a
+  chain-shaped KKT previously constructed a per-supernode ownership map on
+  every factorization before handing off to the sequential driver. Measured
+  at 1.01–1.03× on chain KKTs (clnlbeam, steering_12800, rocket_12800).
 
-### Performance — x86 SIMD kernels actually reach AVX2 (session 2026-08-09)
+### Performance — the dense SIMD kernels now actually vectorize (session 2026-08-09)
 
 - The x86_64 branches of the pulp dispatch helpers now route through
   `Simd::vectorize`, which supplies the `#[target_feature]` context the AVX
@@ -62,10 +74,28 @@ All notable changes to FERAL will be documented in this file.
 - The packed update's pack buffers are pooled in `FactorScratch` on the
   serial path (issue #128 rest): warm factor −12% on a chain-structured KKT
   proxy, −2..6% across the small parity fixtures.
+- Four tuning-knob `env::var` lookups that sat on the **per-panel** dispatch
+  path (one of which also parsed a string every panel) are now read once via
+  `OnceLock`. Measured −5.7% (HYDCAR20) and −9.0% (block-tridiagonal chain)
+  in paired A/B; this is per-front fixed cost, so it lands hardest on the
+  many-small-front matrices where it matters most.
 
-Note: aarch64 (M-series) performance revalidation of the new explicit-SIMD
-packed path is pending; correctness there is guarded by the golden digests,
-and `FERAL_PACKED_SIMD=0` restores the previous behavior if needed.
+aarch64 (M-series) revalidation is **complete and passing** — the packed SIMD
+path was not an x86-only win:
+
+- `bench_dense_front` n=2955 fma serial **2.94×** vs 0.14.0 (11.3 → 34.1
+  GFLOP/s); n=2955 nofma serial 1.55×, n=512 1.37×, n=128 1.13×. Against
+  `FERAL_PACKED_SIMD=0` on the same build the default wins **3.19×**
+  (n=2955) and **2.25×** (n=512), so no per-arch gate or mitigation is
+  needed. On 0.14.0 the aarch64 FMA path was buying nothing (11.3 vs 10.8
+  GFLOP/s); it now delivers.
+- Corpus (156,929 matrices): dense and sparse inertia both **100.0%**, all
+  four Phase 2.8.1 exit partitions PASS, and the worst residuals reproduce
+  *to the digit* against the x86_64 baseline (2.46e-1 `POLAK6_0021`,
+  2.94e-4 `ERRINBAR_0824`).
+- `tests/golden_bits.rs` — digests recorded on x86_64 — passes unmodified on
+  aarch64, so cross-platform bit-identity of the new kernels is demonstrated,
+  not merely argued.
 
 ## [0.14.0] - 2026-07-11
 
--- a/crates/feral-diagnostics/src/bin/diag_chainwoo_profile.rs
+++ b/crates/feral-diagnostics/src/bin/diag_chainwoo_profile.rs
@@ -141,7 +141,7 @@ fn main() {
         total_us_runs.push(total_us);
         prologue_runs.push(report.prologue_us);
         epilogue_runs.push(report.epilogue_us);
-        loop_runs.push(report.loop_us);
+        loop_runs.push(report.loop_us());
 
         if rep == N_REPS - 1 {
             last_timings = p.timings().to_vec();
@@ -176,13 +176,13 @@ fn main() {
 
     // ===== Phase 1b: per-supernode time distribution =====
     println!("\n==== Phase 1b: per-supernode timing distribution ====");
-    let mut us_all: Vec<u64> = last_timings.iter().map(|t| t.us).collect();
+    let mut us_all: Vec<u64> = last_timings.iter().map(|t| t.us()).collect();
     let (us_min, us_p50, us_p90, us_p99, us_max) = percentiles(&mut us_all);
     println!(
         "  per-snode us:  min={}, p50={}, p90={}, p99={}, max={}",
         us_min, us_p50, us_p90, us_p99, us_max
     );
-    let total_loop: u64 = last_timings.iter().map(|t| t.us).sum();
+    let total_loop: u64 = last_timings.iter().map(|t| t.us()).sum();
     println!("  sum per-snode us = {}", total_loop);
 
     // Bucketed by ncol
@@ -190,7 +190,7 @@ fn main() {
     for t in &last_timings {
         let e = by_ncol.entry(t.ncol).or_insert((0, 0));
         e.0 += 1;
-        e.1 += t.us;
+        e.1 += t.us();
     }
     println!("  per-supernode time bucketed by ncol (top by total time):");
     let mut buckets: Vec<(usize, usize, u64)> =
@@ -211,7 +211,7 @@ fn main() {
     for t in &last_timings {
         let e = by_nrow.entry(t.nrow).or_insert((0, 0));
         e.0 += 1;
-        e.1 += t.us;
+        e.1 += t.us();
     }
     println!("  per-supernode time bucketed by nrow (top by total time):");
     let mut buckets: Vec<(usize, usize, u64)> =
@@ -361,7 +361,7 @@ fn main() {
     for t in &last_timings {
         if t.ncol <= 2 {
             tiny_count += 1;
-            tiny_us += t.us;
+            tiny_us += t.us();
         }
     }
     println!(
@@ -428,7 +428,7 @@ fn main() {
         .iter()
         .filter(|t| t.nrow == 32 && t.ncol == 32)
         .collect();
-    top.sort_by_key(|t| std::cmp::Reverse(t.us));
+    top.sort_by_key(|t| std::cmp::Reverse(t.us()));
     for (i, t) in top.iter().take(10).enumerate() {
         let snode = &symbolic.supernodes[t.snode_idx];
         let n_children = snode.children.len();
@@ -445,7 +445,11 @@ fn main() {
             .sum();
         println!(
             "    #{} snode={}: {} us, n_children={}, sum(child trailing rows)={}",
-            i, t.snode_idx, t.us, n_children, total_child_contrib_nrow
+            i,
+            t.snode_idx,
+            t.us(),
+            n_children,
+            total_child_contrib_nrow
         );
     }
 
--- a/crates/feral-diagnostics/src/bin/diag_narx_kernel_gflops.rs
+++ b/crates/feral-diagnostics/src/bin/diag_narx_kernel_gflops.rs
@@ -75,7 +75,7 @@ fn run(iter: usize) {
         (f64::INFINITY, None);
     for t in prof.timings() {
         let f = snode_flops(t.nrow, t.ncol);
-        let us = t.us as f64;
+        let us = t.us() as f64;
         flop_all += f;
         us_all += us;
         if t.nrow > 128 {
@@ -95,7 +95,7 @@ fn run(iter: usize) {
     println!(
         "NARX_CFy_{iter:04}  n={}  loop={:.1} ms  flops={:.3} GFLOP",
         csc.n,
-        report.loop_us as f64 / 1e3,
+        report.loop_us() as f64 / 1e3,
         flop_all / 1e9,
     );
     println!(
@@ -117,7 +117,7 @@ fn run(iter: usize) {
             w.nrow,
             w.ncol,
             snode_flops(w.nrow, w.ncol) / 1e9,
-            w.us as f64 / 1e3,
+            w.us() as f64 / 1e3,
             worst_gflops,
         );
     }
--- a/crates/feral-diagnostics/src/bin/diag_qcqp_profile.rs
+++ b/crates/feral-diagnostics/src/bin/diag_qcqp_profile.rs
@@ -70,7 +70,7 @@ fn run(
         report.n_supernodes,
         factors.factor_nnz(),
         report.total_us,
-        report.loop_us,
+        report.loop_us(),
         report.prologue_us,
         report.epilogue_us,
     );
@@ -85,7 +85,11 @@ fn run(
     for b in &report.buckets {
         println!(
             "  {:<10} {:>8} {:>10} {:>7.1}% {:>8.1}",
-            b.range, b.count, b.sum_us, b.pct_of_total, b.avg_us
+            b.range,
+            b.count,
+            b.sum_us(),
+            b.pct_of_total,
+            b.avg_us()
         );
     }
 
@@ -94,11 +98,11 @@ fn run(
     let mut ncol_us = vec![0u64; NCOL_BUCKETS.len()];
     let mut sum_us_all = 0u64;
     for t in prof.timings() {
-        sum_us_all += t.us;
+        sum_us_all += t.us();
         for (i, &(_, lo, hi)) in NCOL_BUCKETS.iter().enumerate() {
             if t.ncol >= lo && t.ncol <= hi {
                 ncol_counts[i] += 1;
-                ncol_us[i] += t.us;
+                ncol_us[i] += t.us();
                 break;
             }
         }
@@ -127,7 +131,7 @@ fn run(
 
     // Top 5 hottest supernodes
     let mut sorted: Vec<_> = prof.timings().iter().collect();
-    sorted.sort_by_key(|t| std::cmp::Reverse(t.us));
+    sorted.sort_by_key(|t| std::cmp::Reverse(t.us()));
     println!("\ntop 5 hottest supernodes:");
     println!(
         "  {:>5} {:>6} {:>5} {:>5} {:>10}",
@@ -136,13 +140,17 @@ fn run(
     for (rank, t) in sorted.iter().take(5).enumerate() {
         println!(
             "  {:>5} {:>6} {:>5} {:>5} {:>10}",
-            rank, t.snode_idx, t.nrow, t.ncol, t.us
+            rank,
+            t.snode_idx,
+            t.nrow,
+            t.ncol,
+            t.us()
         );
     }
 
     // Cumulative-by-ncol: how much wall time comes from ncol<=K?
     let mut sorted_us_by_ncol: Vec<(usize, u64)> =
-        prof.timings().iter().map(|t| (t.ncol, t.us)).collect();
+        prof.timings().iter().map(|t| (t.ncol, t.us())).collect();
     sorted_us_by_ncol.sort_by_key(|&(c, _)| c);
     let mut cum_count = 0usize;
     let mut cum_us = 0u64;
--- a/crates/feral-diagnostics/src/bin/diag_small_leaf_gate.rs
+++ b/crates/feral-diagnostics/src/bin/diag_small_leaf_gate.rs
@@ -102,7 +102,12 @@ fn main() {
                 let ratio = n.total_us as f64 / o.total_us as f64;
                 println!(
                     "{:>16} | {:>10} {:>10} | {:>10} {:>10} | {:>8.3}",
-                    label, o.total_us, o.loop_us, n.total_us, n.loop_us, ratio
+                    label,
+                    o.total_us,
+                    o.loop_us(),
+                    n.total_us,
+                    n.loop_us(),
+                    ratio
                 );
             }
             _ => println!("{:>16} | run failed", label),
--- a/crates/feral-diagnostics/src/bin/diag_strategy_compare.rs
+++ b/crates/feral-diagnostics/src/bin/diag_strategy_compare.rs
@@ -121,7 +121,14 @@ fn main() {
                 let ratio = r.total_us as f64 / a.total_us as f64;
                 println!(
                     "{:>16} | {:>10} {:>10} {:>10} | {:>10} {:>10} {:>10} | {:>8.3}",
-                    label, a_n, a.total_us, a.loop_us, r_n, r.total_us, r.loop_us, ratio
+                    label,
+                    a_n,
+                    a.total_us,
+                    a.loop_us(),
+                    r_n,
+                    r.total_us,
+                    r.loop_us(),
+                    ratio
                 );
             }
             _ => println!("{:>16} | run failed", label),
--- a/crates/feral-diagnostics/src/bin/probe_narx_phases.rs
+++ b/crates/feral-diagnostics/src/bin/probe_narx_phases.rs
@@ -100,7 +100,7 @@ fn size_distribution(timings: &[SupernodeTiming]) {
         ("65-256", 65, 256),
         (">256", 257, usize::MAX),
     ];
-    let total_us: u64 = timings.iter().map(|t| t.us).sum();
+    let total_us: u64 = timings.iter().map(|t| t.us()).sum();
     println!(
         "    {:>8} {:>7} {:>10} {:>7} {:>9} {:>9}",
         "ncol", "count", "sum_us", "pct", "asm_us", "df_us"
@@ -111,9 +111,9 @@ fn size_distribution(timings: &[SupernodeTiming]) {
         for t in timings {
             if t.ncol >= lo && t.ncol <= hi {
                 count += 1;
-                sum += t.us;
-                asm += t.assembly_us;
-                df += t.densefactor_us;
+                sum += t.us();
+                asm += t.assembly_us();
+                df += t.densefactor_us();
             }
         }
         let pct = if total_us > 0 {
@@ -130,7 +130,7 @@ fn size_distribution(timings: &[SupernodeTiming]) {
 
 fn top_supernodes(timings: &[SupernodeTiming]) {
     let mut v: Vec<&SupernodeTiming> = timings.iter().collect();
-    v.sort_by_key(|t| std::cmp::Reverse(t.us));
+    v.sort_by_key(|t| std::cmp::Reverse(t.us()));
     println!(
         "    {:>6} {:>6} {:>6} {:>9} {:>8} {:>8} {:>8} {:>8} {:>8}",
         "snode", "nrow", "ncol", "us", "asm_us", "df_us", "panel", "schur", "tail"
@@ -141,12 +141,12 @@ fn top_supernodes(timings: &[SupernodeTiming]) {
             t.snode_idx,
             t.nrow,
             t.ncol,
-            t.us,
-            t.assembly_us,
-            t.densefactor_us,
-            t.panelfactor_us,
-            t.schur_us,
-            t.scalartail_us,
+            t.us(),
+            t.assembly_us(),
+            t.densefactor_us(),
+            t.panelfactor_us(),
+            t.schur_us(),
+            t.scalartail_us(),
         );
     }
 }
@@ -186,7 +186,7 @@ fn run(iter: usize) {
     let cold_phases = phase_timing::snapshot();
     let cold_detail = phase_timing::snapshot_detail();
     let cold_zerofill = phase_timing::snapshot_contrib_zerofill();
-    let cold_loop_us = prof.lock().map(|p| p.report().loop_us).unwrap_or(0);
+    let cold_loop_us = prof.lock().map(|p| p.report().loop_us()).unwrap_or(0);
 
     // Warm factor: numeric only (symbolic reused). Re-arm profiler and
     // reset counters.
@@ -224,7 +224,7 @@ fn run(iter: usize) {
     );
     report_loop(
         "WARM",
-        report.loop_us,
+        report.loop_us(),
         warm_phases,
         warm_detail,
         warm_zerofill,
--- a/crates/feral-diagnostics/src/bin/probe_robot_profile.rs
+++ b/crates/feral-diagnostics/src/bin/probe_robot_profile.rs
@@ -53,7 +53,7 @@ fn factor_with_profile(label: &str, csc: &feral::CscMatrix) {
         "  prologue={:.1} ms  epilogue={:.1} ms  loop={:.1} ms  total={:.1} ms  overhead={:.1}%",
         report.prologue_us as f64 / 1e3,
         report.epilogue_us as f64 / 1e3,
-        report.loop_us as f64 / 1e3,
+        report.loop_us() as f64 / 1e3,
         report.total_us as f64 / 1e3,
         report.overhead_pct,
     );
@@ -66,14 +66,14 @@ fn factor_with_profile(label: &str, csc: &feral::CscMatrix) {
             "    nrow {:>7}  {:>6}  {:>8.1}  {:>5.1}%  {:>8.1}",
             b.range,
             b.count,
-            b.sum_us as f64 / 1e3,
+            b.sum_us() as f64 / 1e3,
             b.pct_of_total,
-            b.avg_us,
+            b.avg_us(),
         );
     }
     let top: Vec<_> = {
         let mut v: Vec<_> = prof.timings().iter().collect();
-        v.sort_by_key(|t| std::cmp::Reverse(t.us));
+        v.sort_by_key(|t| std::cmp::Reverse(t.us()));
         v.into_iter().take(10).collect()
     };
     println!("  top 10 supernodes by time:");
@@ -83,7 +83,7 @@ fn factor_with_profile(label: &str, csc: &feral::CscMatrix) {
             t.snode_idx,
             t.nrow,
             t.ncol,
-            t.us as f64 / 1e3,
+            t.us() as f64 / 1e3,
         );
     }
 }
--- a/crates/feral-diagnostics/src/bin/probe_rocket_profile.rs
+++ b/crates/feral-diagnostics/src/bin/probe_rocket_profile.rs
@@ -91,7 +91,7 @@ fn factor_with_profile(label: &str, csc: &feral::CscMatrix) {
         "  prologue={:.1} ms  epilogue={:.1} ms  loop={:.1} ms  total={:.1} ms  overhead={:.1}%",
         report.prologue_us as f64 / 1e3,
         report.epilogue_us as f64 / 1e3,
-        report.loop_us as f64 / 1e3,
+        report.loop_us() as f64 / 1e3,
         report.total_us as f64 / 1e3,
         report.overhead_pct,
     );
@@ -105,14 +105,14 @@ fn factor_with_profile(label: &str, csc: &feral::CscMatrix) {
             "    nrow {:>7}  {:>6}  {:>8.1}  {:>5.1}%  {:>8.1}",
             b.range,
             b.count,
-            b.sum_us as f64 / 1e3,
+            b.sum_us() as f64 / 1e3,
             b.pct_of_total,
-            b.avg_us,
+            b.avg_us(),
         );
     }
     let top: Vec<_> = {
         let mut v: Vec<_> = prof.timings().iter().collect();
-        v.sort_by_key(|t| std::cmp::Reverse(t.us));
+        v.sort_by_key(|t| std::cmp::Reverse(t.us()));
         v.into_iter().take(5).collect()
     };
     println!("  top 5 supernodes by time:");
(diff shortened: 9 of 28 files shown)


jkitchin:
## Context from the pounce side: this is the right fix for #148, and it is *not* the fix for pounce#552 — here is what still needs doing

Reviewed from pounce, where the open item is the factorization gap reported in
[pounce#552](https://github.com/jkitchin/pounce/issues/552): FERAL 3.5–4.8×
slower than MA27/MA57 on chain-structured KKT (hicks N=1200 0.307 s vs 0.092 /
0.088; quad_tank N=1200 0.713 vs 0.167 / 0.161; cart_pole N=600 0.903 vs 0.190 /
0.347). Five notes, roughly in priority order.

### 1. This PR does not move that gap, and its own numbers say so

The diff is scheduling-only — `factor_one_supernode` is called from a new place
but never changed — and `chain12000` goes 12.0–12.5 ms → 11.9–12.5 ms par@4t,
with chains collapsing to one task and taking the new serial fallback. That is
consistent with pounce#552's own stage sweep, which found the parallel dispatch
gate (`feral_min_par_flops` aggressive or off) moved factorization time only
10–20%, in inconsistent directions, against a 3.5–4.8× gap. Nothing wrong with
the PR — it fixes #148, which is a different problem. Flagging it only because
"parallel driver perf work landed" reads from outside like the chain gap was
addressed.

### 2. The highest-value next action is a release, not more code

The 2026-08-09 SIMD work already on `main` is, as far as I can tell, the actual
answer to pounce#552:

- the x86_64 pulp dispatch helpers were running lane ops as outlined calls
  (~10× kernel slowdown, invisible on aarch64);
- the packed BLAS-3 trailing update's tile walk compiled **fully scalar** on x86
  (objdump: ymm=0, 187 scalar `mulsd`/`subsd`, ~6 GF/s scalar-ILP ceiling);
- pack-buffer pooling gives −12% on `chain1200` — the hicks-like
  block-tridiagonal synthetic that was added *because of* the pounce#552 report.

A fully scalar inner kernel against MA57's BLAS3 is a sufficient explanation for
3.5–4.8× on its own, and it matches what pounce#552 concluded by elimination
("the cost appears to sit in the numeric kernel rather than in a mis-selected
heuristic").

But **none of it has reached pounce**: pounce pins `feral = "0.14.0"` (released
2026-07-11), and this is all unreleased `main`. So the 3.5–4.8× figure still
describes the pre-SIMD kernel, and it will keep describing it until there is a
0.15.0. Cutting that release — and then re-running the hicks / quad_tank /
cart_pole factorization comparison against MA27/MA57 — is worth more right now
than any further optimization, because it is also the measurement that tells you
how much of the gap is left.

### 3. The chainW anomaly deserves an issue, not a research-note footnote

The documented open question in this PR is that the old per-node-spawn driver
beat *both* sequential drivers on chainW, with telemetry placing the difference
inside `factor_one_supernode` (912 vs 1276 ms over 9 identical factors) — not
spawn, not lock, not driver, not intrafront.

Two reasons that is more interesting than "synthetic-proxy quirk":

- It is the same place pounce#552 pointed at by elimination, on the same tree
  shape, at a similar magnitude (~40%).
- chainW is a **wide-block** chain, and that is not a curiosity — it is exactly
  the shape pounce#552 names as the case that will bite: "it would become one for
  a model that combines many time points with a large per-time-point block."
  That is `double_column` (n = 25,377, 31 time points, 819 vars/point) and
  `prommis_sx` (n = 28,313, 913 vars/point), the two models pounce is currently
  *winning* on. Losing 40% inside the kernel on that shape is a live risk to a
  live result, not a proxy artifact.

Testable hypotheses that need neither `perf` nor `heaptrack`, since the fast and
slow runs execute identical nodes and differ only in caller context:

- **Workspace shape churn.** The serial driver walks one shared `FactorScratch`
  through a chain of differently-shaped fronts, resizing up and down; the old
  per-node driver handed each node a workspace from the thread pool, so buffers
  stayed in a narrower size class. Instrument `FactorScratch` resize/realloc
  counts across a chainW factorization on both drivers — a large asymmetry
  confirms it, and a round-robin pool of N workspaces on the serial path is the
  cheap experiment.
- **Pool coverage asymmetry.** The 2026-08-09 decision put `PackPool` on the
  serial packed path only, with the intra-front rayon path keeping per-range
  fresh allocations. chainW's direction (parallel faster) is the opposite of what
  pooling predicts, so an A/B of the new serial driver with the pool disabled
  would say whether the pool is implicated at all — and if disabling it *helps*
  on chainW, that is the finding.
- **First-touch locality.** Workspace pages faulted in on a worker thread vs the
  main thread. Pinning the sequential driver's factor body onto a rayon worker
  (one-thread pool) isolates this from everything else.

### 4. aarch64 is unmeasured, and pounce ships M-series wheels

The changelog notes M-series revalidation of the explicit-SIMD packed path is
pending, with `FERAL_PACKED_SIMD=0` as the mitigation. pounce publishes macOS
arm64 wheels, so that revalidation is on the critical path for a release, not
after it. Worth having a number before 0.15.0 goes out.

### 5. Smaller, already on your list

- #148 suggestion 3 (the 2.9M `collect()` temporaries) — you deferred it pending
  re-profiling after coarsening; that re-profile is now unblocked.
- The #128 `nrow` underestimate is harmless for the gate as documented, but it
  also skews `subtree_flops`, which is now load-bearing for the task-boundary
  decision, not just for a yes/no dispatch. Might be worth re-checking the
  sibling rule's sensitivity to it.

One thing FERAL cannot self-serve: pounce#552's remaining ask is a like-for-like
HSL comparison, which needs an HSL-linked build that neither pounce CI nor the
reporter's environment has. A documented GF/s figure for the dense front at a
couple of sizes would be the next best thing — it lets the kernel be judged
against a BLAS3 reference without anyone needing HSL.

---

jkitchin:
## Investigated the chainW question — the anomaly was measurement noise, and I'm withdrawing it

You were right to push on this, and the push was worth it: it produced a correction that runs deeper than the anomaly itself. The answer is not the one either of us expected.

### The anomaly does not exist

Decisive bisect first — run the *new* driver across `FERAL_PAR_TASK_MIN_FLOPS` cutoffs on chainW:

```
cutoff=1          n_tasks=18717  seeds=17965   median_us=166056
cutoff=1000       n_tasks=5      seeds=3       median_us=146261
cutoff=100000     n_tasks=1      seeds=1       median_us=139582
cutoff=1000000    n_tasks=1      seeds=1       median_us=259086   <--
cutoff=100000000  n_tasks=1      seeds=1       median_us=155524
```

The last three produce **identical task plans** — `n_tasks=1, seeds=1`, literally the same code path executing byte-identical work — and span 139.6 → 259.1 ms. A 1.9× spread on identical code. Eight invocations of one fixed config confirm it: `min_us` 124.7–163.2 ms (31%), `median_us` 149.2–183.7 ms (23%).

Both conclusions previously drawn from inside that band were wrong: my "chainW anomaly" (per-node spawning ~20% faster) **and** the 5–18% regression it implied.

### Paired re-measurement reverses the sign

Alternating A/B within one time window so slow drift cancels; `min_us` per invocation; ratio < 1 means new is faster:

| comparison | pairs favouring new/coarse | median ratio | reading |
|---|---|---:|---|
| chainW — new (PR150) vs old (main) | 9 / 12 | **0.961** | new ~4% **faster** |
| chainW — coarse vs fine-grained, *same binary* | 9 / 10 (p≈0.02) | **1.045** | per-node tasks ~4% **slower** |
| sparseqpL — new vs old | 9 / 10 (p≈0.02) | 0.894 | new ~11% faster |
| grid250 — new vs old | **10 / 10** (p≈0.002) | 0.762 | new ~24% faster |

So on the wide-block chain, the shipped coarsening is mildly *better* than per-node spawning — the reverse of the withdrawn anomaly — and there is no regression on the shape you flagged. The `AtomicLockStats` reading I cited (`factor_body` 912 vs 1276 ms) was also unpaired; **withdrawn, please don't cite it**. Row 2 is the load-bearing one: same binary, same input, only granularity differs, so it isolates the effect you were worried about.

I also re-ran the PR's own headline claims paired, since they were collected the same flawed way — they survive and strengthen (grid250 went from "−9…−21%" to a clean 10/10 at −24%).

### The methodology finding is the real output

The project rule was "any bench delta below ~5% needs a 3-run median" (tried-and-rejected 2026-04-14). That is **not sufficient** here: three consecutive medians can all land inside one drift excursion, which is exactly how both wrong conclusions were reached. New rule recorded in `decisions.md`: perf claims in this container require paired alternating A/B, ≥10 pairs, sign test, `min_us` per sample; cross-session number comparison is not evidence at all. I've also flagged that sub-10% fixture deltas in the two 2026-08-09 checkpoints should be treated as unresolved rather than as measured wins until re-run paired.

Research note corrected in 94a90ad; the PR body's caveat section is now wrong and I'll update it.

### On your other four points

1. **Release before more code** — agreed, and it's the top item. pounce pins 0.14.0, so the 3.5–4.8× figure still describes the pre-SIMD kernel (no x86 dispatch fix, no AVX2 packed kernel). A fully scalar inner kernel against MA57's BLAS3 is a sufficient explanation on its own.
2. **aarch64 gates the release, not follows it** — agreed, and it's the one thing I can't do from here (x86 container). Happy to prepare the exact command list and pass/fail criteria.
3. **GF/s instead of an HSL build** — cheap and I have the numbers: dense front n=2955 at **5.6 GF/s** serial nofma and **14–15 GF/s** intrafront+FMA on a 4-core AVX2 container, against ~6 GF/s scalar-ILP before. I'll write them up at a few sizes as a documented reference point.
4. **`nrow` underestimate now load-bearing for task boundaries** — fair hit, and sharper than my note admitted: it graduated from a yes/no dispatch gate to deciding *where* boundaries fall. Worth a sensitivity check on the sibling rule.

---

jkitchin:
## Correction to my earlier comment — plus the MA57 head-to-head, which is now measurable

Two things. First, point 2 of my comment above was wrong and I am retracting it.
Second, I found a CoinHSL build for this arch, so the "someone needs an
HSL-linked build" caveat no longer applies — there are now direct
FERAL-vs-MA57 factorization numbers at the bottom, which is the measurement
pounce#552 has been missing since it was filed.

### 1. Retraction: the SIMD work is not the pounce#552 fix

I claimed the 2026-08-09 kernel work on `main` was "as far as I can tell, the
actual answer to pounce#552". I built pounce against the git rev and measured.
It is not. Your own session note already said so and I had not read it:

> the chain-KKT (pounce#552) gap needs the small-front *overhead* work, not lane
> width (see rejection entry): per-front fixed overhead, pivot scans,
> `scalar_pivot_step`, delayed-pivot cascade are the ranked suspects.

That is the correct call, and the data below supports it over my version.

**pounce against `feral @ e8e1c5a` vs the pinned `feral 0.14.0`.** Uncommitted
`[patch.crates-io]` redirect, two release binaries (verified distinct — one links
the `feral-0.14.0` registry path, the other carries the `FERAL_PACKED_SIMD`
strings). Eight models, three interleaved reps, median total wallclock. Apple
M-series.

| model | 0.14.0 | main | Δ |
|---|---:|---:|---:|
| optcontrol | 0.863 s | 0.874 s | +1.3% |
| dtoc2 | 2.713 | 2.710 | −0.1% |
| dtoc1nd | 0.957 | 0.942 | −1.5% |
| clnlbeam | 7.516 | 7.587 | +0.9% |
| steering_12800 | 8.654 | 8.585 | −0.8% |
| rocket_12800 | 11.362 | 11.488 | +1.1% |
| marine_1600 | 2.362 | 2.354 | −0.3% |
| poisson | 2.280 | 2.280 | −0.0% |

Iteration count, `n_factors`, `last_nnz_l` and final objective match exactly
between binaries on every model, so both walked the identical path. A third arm
(new binary, `FERAL_PACKED_SIMD=0`) is flat too.

### 2. aarch64 revalidation: no regression, and a large win

Item 1 of your "Next Session Should", and the gap my point 4 flagged.
`bench_dense_front`, medians of 3, `c05eb77` vs `e8e1c5a`:

| case | 0.14.0 | main | speedup |
|---|---:|---:|---:|
| n=2955 nofma serial | 775–809 ms (10.6 GF/s) | 500–509 ms (17.2 GF/s) | 1.55× |
| n=2955 nofma intrafront | 140–143 ms (60.8 GF/s) | 96 ms (89.9 GF/s) | 1.47× |
| n=2955 fma serial | 755–759 ms (11.3 GF/s) | 252–261 ms (34.1 GF/s) | **2.94×** |
| n=2955 fma intrafront | 127–132 ms (66.7 GF/s) | 71–73 ms (120.5 GF/s) | 1.79× |
| n=512 nofma serial | 4.19–4.34 ms (10.5 GF/s) | 3.07–3.11 ms (14.5 GF/s) | 1.37× |
| n=256 nofma serial | 0.56 ms (10.0 GF/s) | 0.44 ms (12.8 GF/s) | 1.27× |
| n=128 nofma serial | 0.080 ms (8.5 GF/s) | 0.070 ms (9.6 GF/s) | 1.13× |

`FERAL_PACKED_SIMD=0` on `main` reverts to 857–862 ms — at or slightly below
0.14.0 — so the win is attributable to the packed SIMD kernel and the escape
hatch works. Note the FMA path in particular: on 0.14.0 aarch64, `fma serial`
was 11.3 GF/s against `nofma serial` at 10.8, i.e. FMA was buying nothing. On
`main` it is 34.1 vs 17.2. The scalar tile-walk problem was not purely an x86
story — NEON was leaving ~3× on the floor at large fronts.

Gates green on aarch64: 409 lib tests, `golden_bits` passing under default /
`FERAL_PACKED_SIMD=0` / `FERAL_PACKED_SCHUR=0`, full `cargo test --release`
exit 0. Those digests were recorded on x86 and pass unmodified here — a nice
cross-arch bit-exactness result for the new kernel. **This unblocks the release
rather than gating it.**

One caveat on the bench set: `bench_schur_micro` is **not** informative for this
change. Its packed path is 19.3–19.7 ms in all three configurations (0.14.0,
main, main with SIMD off) — it reaches an entry point the PR did not touch. Only
its strided baseline moves, and that is noise. If it is meant to cover the new
kernel, it currently doesn't.

The two results reconcile cleanly: the speedup is monotone in front size (1.13×
at n=128 → 2.94× at n=2955), and chain-structured KKT is many small fronts.
Different regimes, exactly as your rejection entry argues.

### 3. The measurement that was missing: FERAL vs MA57, factorization only

pounce has had an `ma57` feature and a `pounce-hsl` backend all along; what was
missing was an HSL build. There is a CoinHSL 2023.11.17 aarch64 binary
distribution, so I built `pounce-cli --features ma57` against it and wrote a
factorization-only harness: dump a KKT matrix out of a real solve
(`POUNCE_DBG_KKT_DUMP`, iteration 3), then feed the identical triplet to both
backends through `SparseSymLinearSolverInterface` and time `multi_solve`.

Medians of 15, `POUNCE_FERAL_REFINE=0`, MA57 pinned to one thread
(`OPENBLAS_NUM_THREADS=1`), FERAL left rayon-parallel. Apple M-series.

| matrix | dim | nnz | nnz/dim | FERAL | MA57 | ratio |
|---|---:|---:|---:|---:|---:|---:|
| clnlbeam | 99,999 | 299,993 | 3.0 | 57.80 ms | 7.18 ms | **8.05×** |
| dtoc1nd | 9,685 | 223,230 | 23.0 | 16.84 | 3.47 | 4.85× |
| marine_1600 | 76,807 | 414,583 | 5.4 | 39.38 | 12.92 | 3.05× |
| rocket_12800 | 89,601 | 473,589 | 5.3 | 24.70 | 8.62 | 2.87× |
| steering_12800 | 115,201 | 422,392 | 3.7 | 47.34 | 18.68 | 2.53× |
| dtoc2 | 103,920 | 2,051,790 | 19.7 | 139.84 | 82.50 | 1.70× |

Back-substitution only, same runs: FERAL 0.84–9.08 ms vs MA57 0.30–3.40 ms,
i.e. 1.4–2.8×.

`number_of_neg_evals` is **identical** between the two backends on every matrix
(40000, 3726, 38392, 38402, 51201, 38970), so this is genuinely the same problem
solved the same way — not two different pivot sequences.

Three things I would draw from it:

- **It reproduces pounce#552 independently**, on different models, with the
  confound removed. 1.7–8.1× against the reported 3.5–4.8×. The gap is real and
  it is in the factorization, not in pounce.
- **The ordering matches the overhead hypothesis, not the throughput one.** The
  worst case is `clnlbeam` at 8.05×, which is also the *sparsest* matrix in the
  set (nnz/dim = 3.0 — essentially a pure chain). The best case is `dtoc2` at
  1.70×, which is by far the *densest* (nnz/dim = 19.7, big fronts). That is the
  signature of per-front fixed cost, not of kernel GF/s. `dtoc1nd` is the
  exception to the trend, and it is also the smallest matrix by an order of
  magnitude, where fixed cost per front is again proportionally largest. If you
  want one fixture to drive the overhead work, `clnlbeam` at 8× is the sharpest
  one here, and it is a plain checked-in Mittelmann model.
- **FERAL is losing this while multithreaded.** MA57 is pinned to one thread;
  FERAL had the full rayon pool. Per core the gap is wider than the table shows.

Fairness notes, both directions. FERAL's iterative refinement is **on by
default** in pounce (`FeralConfig::from_env`, `refine: true`), and MA57's
backend does not refine (upstream Ipopt does it at the driver level). With
refinement on — the configuration pounce actually ships — the clnlbeam numbers
are 82.91 ms vs 7.24 ms (11.4×) and backsolve 17.65 ms vs 1.56 ms (11.3×). I
used refine-off above because that is the like-for-like comparison, but the
shipped default is the worse number. Conversely, the refinement is not
gratuitous: with it off, FERAL's relative residual on `marine_1600` is 8.1e-7
against MA57's 2.9e-10, so it is buying real accuracy on some matrices. Where
FERAL does refine it is consistently the more accurate of the two (1e-13…1e-15
vs MA57's 1e-8…1e-9 on several of these).

### 4. One incidental find, in the regime that matters

`apply_schur_panel_range_packed_impl` reads
`std::env::var("FERAL_PACKED_SIMD_MIN_WORK")` on **every panel**, to recover a
compile-time constant:

```rust
let min_work = std::env::var("FERAL_PACKED_SIMD_MIN_WORK")
    .ok()
    .and_then(|v| v.parse::<usize>().ok())
    .unwrap_or(PACKED_SIMD_MIN_WORK);
```

`packed_schur_enabled()` (from `apply_blocked_schur` and
`apply_schur_panel_range`) and `packed_simd_enabled()` are uncached the same way
— there is no `OnceLock`/`LazyLock` anywhere in `src/dense/factor.rs`. Each call
heap-allocates a `String`, and on macOS `__findenv_locked` takes a process-global
lock that the intra-front rayon path will contend on.

Measured, it is small: padding the environment with 400 extra variables costs
~1–2% at n=128 and ~0.4% at n=256. I flag it anyway because it is *per-front
fixed overhead*, which is precisely the category your note ranks as the #552
suspect, and hoisting all three into a `OnceLock` is a few lines with no
behaviour change.

---

jkitchin:
## aarch64 revalidation for 0.15.0, and the `FERAL_PAR_MIN_SEEDS` answer: **keep it at 2**

Ran the release checklist on the M-series box, plus the calibration you asked
for. Headline: **the correctness gate passes**, **the knob should not be
changed** — and the reason it should not be changed is the interesting part.
All numbers below are `6fd12d4`, Apple M-series (14 rayon threads), paired
alternating A/B with a sign test, per `dev/decisions.md` 2026-08-09.

### 1a. Correctness gate — PASS

```
cargo test --release
84/84 test binaries · 802 passed · 0 failed · 21 ignored
tests/golden_bits.rs: 3 passed
  golden_bits_fma_indefinite_256    ok
  golden_bits_nofma_indefinite_256  ok
  golden_bits_nofma_2x2_heavy_96    ok
```

The x86_64-recorded `to_bits` digests reproduce on aarch64. Cross-platform
bit-identity of the new SIMD kernels is proven, so that gate is cleared.

Corpus run (`cargo run --bin bench --release`) is still going — multi-hour on
the 22 GB corpus. I'll post inertia / Phase 2.8.1 partitions separately; it
does not block the rest of this.

### 2. `FERAL_PAR_MIN_SEEDS`: the answer is 2, i.e. don't change `PAR_MIN_SEEDS`

You asked me to sweep it and tell you a number. The number is the one you
already shipped, and it isn't a coincidence — **the coarsening in 7814bfe3
already solved the problem the knob was built to work around.**

`FERAL_DEBUG_TASK_PLAN=1` on the six matrices:

| matrix | n_snodes | n_tasks | seeds | falls back at default? |
|---|---:|---:|---:|---|
| clnlbeam | 39,582 | **1** | **1** | yes |
| dtoc1nd | 3,831 | **1** | **1** | yes |
| steering_12800 | 19,197 | **1** | **1** | yes |
| rocket_12800 | 5,125 | **1** | **1** | yes |
| dtoc2 | 29,885 | **1** | **1** | yes |
| marine_1600 | 8,809 | 3 | 2 | no |

Five of six collapse to a *single task*. `seeds=1 < 2`, so they already take
the serial fallback at the shipped default, and the threshold is a no-op on
them at every setting. That is why my #148 "sequential beats parallel on 4–5
of 6" finding no longer reproduces: there is no longer a spawn-storm to lose
to.

`marine_1600` is the only matrix where the knob does anything — and there it
is a cliff in the direction that says *keep 2*:

| arm | min ms | med ms | vs default | sign wins | p |
|---|---:|---:|---:|---:|---:|
| par ms=1 | 30.87 | 31.42 | 1.00× | 9/15 | 0.61 |
| **par ms=2 (default)** | **31.04** | **31.34** | — | — | — |
| par ms=4 | 39.57 | 40.33 | 0.78× | 0/15 | 0.0001 |
| par ms=8 | 39.73 | 40.37 | 0.78× | 0/15 | 0.0001 |
| par ms=16 | 39.41 | 40.08 | 0.78× | 0/15 | 0.0001 |
| par ms=64 | 39.68 | 40.13 | 0.78× | 0/15 | 0.0001 |
| serial driver | 39.28 | 39.95 | 0.78× | 0/15 | 0.0001 |

Raising the default to 4 or beyond costs **28%** on the one matrix in my
corpus with real tree parallelism and buys nothing on the other five. So the
conservative default is correct, and it is correct for a better reason than
when you picked it.

On the other five, every setting **at or above** 2 is inside ±1% and mostly not
significant — clnlbeam ms=4 1.01× (11/15, p=0.12), ms=8 1.00× (7/15, p=1.0),
ms=64 1.01× (10/15, p=0.30), and similarly for the rest. The one setting that
does move them is `ms=1`, which disables the fallback and forces the task graph
even at a single task:

| matrix | ms=1 vs default | wins | p |
|---|---:|---:|---:|
| clnlbeam | 0.91× | 0/15 | 0.0001 |
| steering_12800 | 0.95× | 0/15 | 0.0001 |
| rocket_12800 | 0.96× | 0/15 | 0.0001 |
| dtoc2 | 0.97× | 1/15 | 0.001 |
| dtoc1nd | 0.97× | 3/15 | 0.035 |

That is a clean measurement of what the serial fallback itself is worth once
coarsening has already collapsed the tree: **3–9%**, significant on all five.
So both halves of 7814bfe3 are earning their keep, separately.

**Parity confirmed independently.** I hashed the solve vector at all seven
settings on all six matrices: bit-identical everywhere. That's `task_plan_parity.rs`'s
claim reproduced on real KKTs on a second architecture.

### 3. What #150 actually buys — 0.14.0 vs `6fd12d4`, paired

Two binaries, one linking the crates.io `feral-0.14.0`, one the #150 worktree
(verified distinct: `FERAL_PAR_MIN_SEEDS` appears in one and not the other).
They can't be linked into one process, so I alternated *process launches* —
A,B,A,B — 12 pairs per cell, `min` of 5 warm factorizations per launch.

| matrix | driver | 0.14.0 | #150 | speedup | wins | p |
|---|---|---:|---:|---:|---:|---:|
| clnlbeam | serial | 23.30 | 22.02 | 1.06× | 11/12 | 0.006 |
| clnlbeam | parallel | 53.07 | 23.18 | **2.29×** | 12/12 | 0.0005 |
| dtoc1nd | serial | 12.94 | 12.15 | 1.06× | 11/12 | 0.006 |
| dtoc1nd | parallel | 13.73 | 12.02 | 1.14× | 11/12 | 0.006 |
| steering_12800 | serial | 26.11 | 22.44 | 1.16× | 12/12 | 0.0005 |
| steering_12800 | parallel | 39.11 | 22.59 | **1.73×** | 12/12 | 0.0005 |
| rocket_12800 | serial | 15.85 | 15.58 | 1.02× | 10/12 | 0.039 |
| rocket_12800 | parallel | 18.11 | 15.65 | 1.16× | 12/12 | 0.0005 |
| marine_1600 | serial | 38.27 | 36.94 | 1.04× | 12/12 | 0.0005 |
| marine_1600 | parallel | 29.10 | 28.73 | 1.01× | 8/12 | 0.39 |
| dtoc2 | serial | 91.16 | 89.25 | 1.02× | 11/12 | 0.006 |
| dtoc2 | parallel | 112.42 | 96.70 | 1.16× | 12/12 | 0.0005 |

#150 wins all twelve arms. The big ones are exactly where the coarsening bites:
the default-pool penalty I reported in #148 (clnlbeam 1.99× *slower* threaded)
is now 1.05×, and steering's is gone. **The solve-vector hash is identical
between 0.14.0 and #150 on all six matrices** — so this entire delta, SIMD
kernels included, is bit-neutral on real KKT.

### 4. One residual: the serial fallback isn't free

`build_task_plan(symbolic)` is called at `factorize.rs:3148`, one line *before*
the `plan.seeds.len() < par_min_seeds()` check at :3149. So a chain KKT builds
a complete task plan over every supernode and then throws it away and calls the
sequential driver — on every factorization, since the plan is rebuilt each time.

It is measurable. Serial driver vs the parallel driver's own fallback path,
15 paired rounds:

| matrix | serial | par (falls back) | penalty | wins | p |
|---|---:|---:|---:|---:|---:|
| clnlbeam | 26.06 | 26.91 | 1.03× | 14/15 | 0.001 |
| steering_12800 | 23.93 | 24.28 | 1.01× | 13/15 | 0.007 |
| rocket_12800 | 16.81 | 16.95 | 1.01× | 12/15 | 0.035 |
| dtoc2 | 100.88 | 101.37 | 1.00× | 10/15 | 0.30 |
| dtoc1nd | 13.55 | 13.63 | 1.01× | 9/15 | 0.61 |

Only the top three clear significance; dtoc2 and dtoc1nd are consistent in sign
but I would not claim them. Small, but consistently signed and cheap to remove: the plan is a pure function
of `symbolic`, which is fixed across refactorizations, so it can be cached
there — or the seed count can be computed before committing to building
`owned`/`parent_task`/`task_pending_init`. Note `owned: Vec<Vec<usize>>` is a
`Vec` per supernode, so clnlbeam allocates ~39.6k short-lived `Vec`s per
factorization to produce a plan it discards. (Same shape as the pounce-side bug
in pounce#562, which I fixed this morning — rebuilding a fixed structure every
IPM iteration.) Not a blocker for 0.15.0.

### 5. Item (2) verified on real matrices — and it re-ranks item (4)

The nanosecond profiler reproduces your result and then some. **Zero supernodes
record a literal 0 ns on any of the six.** More usefully, the population that
was invisible before is now sized:

| matrix | ≤8 fronts | ≤8 avg | ≤8 share of loop | dominant bucket |
|---|---:|---:|---:|---|
| clnlbeam | 35,649 (90%) | 0.150 µs | **35.9%** | 9–16 @ 39.6% |
| steering_12800 | 12,797 (67%) | 0.206 µs | **14.6%** | 9–16 @ 85.4% |
| marine_1600 | 1,414 (16%) | 0.390 µs | 1.7% | 33–64 @ 58.8% |
| rocket_12800 | 5 (0.1%) | 1.758 µs | 0.1% | 17–32 @ 88.7% |
| dtoc1nd | 25 (0.7%) | 0.415 µs | 0.1% | 33–64 @ 98.6% |
| dtoc2 | 29 (0.1%) | 0.460 µs | 0.0% | 65–128 @ 51.1% |

On `clnlbeam` the ≤8 bucket is 90% of all supernodes and **over a third of the
loop** — and under the old `as_micros()` field every one of those 35,649 fronts
read as costless. That is the single most under-reported thing in the old
profile, and it is now the best-evidenced target you have.

But it re-ranks item (4) rather than promoting it globally: `nemin`
amalgamation is a **clnlbeam/steering fix**, worth 36% / 15% of their loops,
and worth essentially nothing on dtoc1nd, rocket or dtoc2, whose loops sit in
33–64 and 65–128 column fronts. This is the same split as the `dtoc1nd`
counterexample I raised on #148 — "chain KKT is many small fronts" is true of
some chain KKTs, not the class.

### 6. Item (3) scaling: confirmed the top prologue item, and confirmed it does not warm

The third (warm) factorization, serial, profiled:

| matrix | total | prologue | scaling | scaling % of prologue | scaling % of total |
|---|---:|---:|---:|---:|---:|
| clnlbeam | 21.74 ms | 5.17 | 3.91 | 76% | 18% |
| dtoc1nd | 12.83 | 3.06 | 2.33 | 76% | 18% |
| steering_12800 | 22.77 | 3.94 | 2.25 | 57% | 10% |
| rocket_12800 | 17.46 | 7.66 | 5.95 | 78% | **34%** |
| marine_1600 | 38.60 | 6.28 | 5.11 | 81% | 13% |
| dtoc2 | 94.36 | 25.74 | 19.85 | 77% | **21%** |

The clean part of this: across calls 0→2, `permute` collapses (clnlbeam
4443→539 µs, dtoc2 25647→1584 µs) because pattern reuse works, while `scaling`
stays flat (3909 / 19850 µs). So the prologue *does* warm — for everything
except the one component that dominates it. On `rocket_12800` scaling alone is
34% of the entire factorization. Item (3) is correctly ranked as the largest
remaining line item.

### 7. 1b — NEON SIMD A/B: the default wins, no mitigation needed

The open question was whether the explicit-SIMD packed kernel beats the old
autovectorized tile walk on NEON, where LLVM might already have vectorized it.
It does not — LLVM did not, and the packed kernel wins by more than it did on
x86. 12 pairs, `fma serial` arm:

| fixture | default | `FERAL_PACKED_SIMD=0` | speedup | wins | p |
|---|---:|---:|---:|---:|---:|
| `bench_dense_front 2955 5` | 254.48 ms | 812.76 ms | **3.19×** | 12/12 | 0.0005 |
| `bench_dense_front 512 20` | 1.98 ms | 4.45 ms | **2.25×** | 12/12 | 0.0005 |

`bench_schur_micro 2048 2048 64 10`: packed 35.57 ms vs strided baseline 53.38 ms,
**1.50×**, `byte-exact: baseline == packed ✓`. Unaffected by `FERAL_PACKED_SIMD`
(that gate gates the dense-front path, not this one).

So: ship the default, no `cfg(target_arch)` gate, no retune of
`FERAL_PACKED_SIMD_MIN_WORK` — nothing regressed at n=512 either.

### Summary against your checklist

- **1a correctness** — PASS (84/84, golden_bits 3/3). Corpus in flight.
- **1b performance** — default wins 2.25–3.19× on NEON. Not a blocker, and not
  a problem.
- **2 `FERAL_PAR_MIN_SEEDS`** — keep `PAR_MIN_SEEDS = 2`. Recommend recording
  *why* in the doc comment: the coarsening leaves chain KKTs at `seeds=1`, so
  the threshold only ever decides for trees that have real parallelism, and for
  those a low value is right. The current comment says the default is "known to
  be too low" on the strength of my #148 data — that is no longer true after
  7814bfe3 and is worth correcting before release.
- **Post-release order** — I'd keep yours (scaling warm-start → nemin →
  permute-cache), with the note that nemin is now known to be a two-matrix win
  rather than a general one, and with §4 above as a cheap extra.

Raw output: `seedbench` (7-arm in-process sweep) and the cross-version pair log
are reproducible from the KKT dumps via `POUNCE_DBG_KKT_DUMP`; happy to attach
either if useful.

---

jkitchin:
## Corpus run finished — **all MUST gates pass on aarch64**. 0.15.0 is clear to ship.

Completing the one item left open in my comment above. `cargo run --bin bench
--release` at `6fd12d4`, Apple M-series, full 22 GB corpus — **156,929
matrices**.

### Inertia — 100% on both paths

| path | inertia match | residual pass | worst residual |
|---|---|---|---|
| dense (n ≤ 1000) | **154,429 / 154,482 (100.0%)** | 154,149 / 154,482 (99.8%) | 2.46e-1 `POLAK6_0021` |
| sparse vs MUMPS | **154,531 / 154,590 (100.0%)** | 154,258 / 154,590 (99.8%) | 2.94e-4 `ERRINBAR_0824` |

The worst residuals are the interesting bit: the checklist names the 2026-07-11
baseline as "dense 2.46e-1 POLAK6_0021, sparse 2.94e-4 ERRINBAR_0824". Both
reproduce **to the digit, on the same matrices, on a different architecture.**
Not "not materially worse" — identical. Combined with `golden_bits` passing and
the solve-vector hashes matching 0.14.0 on all six pounce KKTs, that is three
independent lines of evidence that the 2026-08-09 kernel work is bit-neutral
across x86_64 and aarch64.

(2,339 matrices are parse-skipped for `non-finite value 'inf'` in the `.mtx` —
CERI651A/B, SPANHYD and friends. Pre-existing corpus data issue, unrelated to
this branch.)

### Phase 2.8.1 exit partitions — 4/4 PASS

| path | bucket | count | p90 | target | verdict |
|---|---|---:|---:|---|---|
| dense | small-frontal (<200) | 147,982 | 1.59 | ≤ 2.0 | **PASS** |
| dense | medium (<500) | 152,145 | 1.96 | ≤ 3.0 | **PASS** |
| sparse | small-frontal (<200) | 153,455 | 1.56 | ≤ 2.0 | **PASS** |
| sparse | medium (<500) | 153,560 | 1.56 | ≤ 3.0 | **PASS** |

### vs the oracles, for the release note

| ratio | geomean | p50 | p90 | p99 |
|---|---:|---:|---:|---:|
| sparse factor / MUMPS | 0.44 | 0.30 | 1.56 | 3.33 |
| sparse solve / MUMPS | 0.07 | 0.08 | 0.14 | 0.66 |
| sparse factor / SSIDS | 0.04 | 0.03 | 0.32 | 0.96 |
| sparse nnzL / MUMPS | 0.61 | 0.58 | 0.75 | 4.50 |

Sparse factor beats MUMPS at the geomean and the median and only crosses 1.0
in the p90 tail, with a max of 10.12× — against a dense-path max of 94.72×.
Fill is also lower than MUMPS at every quantile below p99.

Worth stating plainly for the release note, since it is the counterpoint to
pounce#552: on this corpus FERAL is *not* generally slower than the reference
solvers — the chain-structured KKT that pounce feeds it is a tail case, not the
typical case. Which is consistent with what I found on the six: the pathology
was the parallel driver's task granularity, and #150 fixes it.

### Checklist status

- **1a correctness** — PASS. `cargo test --release` 84/84 binaries, 802 passed
  / 0 failed; `golden_bits` 3/3; corpus inertia 100% both paths; all four
  partitions PASS; worst residuals identical to the x86_64 baseline.
- **1b performance** — default packed SIMD wins 2.25–3.19× on NEON. No
  mitigation, no `cfg(target_arch)` gate, no `FERAL_PACKED_SIMD_MIN_WORK`
  retune needed.
- **2 `FERAL_PAR_MIN_SEEDS`** — keep `PAR_MIN_SEEDS = 2` (see previous comment
  for why; the doc comment calling it "known to be too low" wants updating).

Nothing here blocks the release. From the pounce side the thing I want most is
0.15.0 on crates.io so I can drop the `[patch.crates-io]` redirect and re-run
pounce#552's headline models against a pinned version.

---

jkitchin:
## All four action items addressed in 0e86167 — and thank you for the retraction, it settles the strategy

Three of your findings landed as code, one as documentation. Taking them in the order they matter.

### 1. `PAR_MIN_SEEDS` doc corrected — your calibration reversed my justification

You asked for this before release and you're right. The comment claimed the default was "known to be too low" on the strength of your #148 data; that data is superseded by your own §2. The doc now records the actual reason it should stay at 2:

> coarsening collapses five of the six to a **single task** (`seeds = 1`), so they already take the sequential fallback and the threshold is a no-op on them at every setting. It therefore only ever decides for trees that *do* have real parallelism — and for those a low value is right: on `marine_1600`, raising it to 4 or beyond costs **28%** (0.78×, 0/15 pairs, p = 1e-4).

That's a better argument than the one I shipped it with. The fallback branch also still cited the withdrawn chainW anomaly — replaced with your measurement that forcing the task graph on (`ms=1`) costs 3–9% across the five.

### 2. §4 — the discarded task plan, fixed

`build_task_plan` now returns `Option<TaskPlan>` and takes `min_seeds`, returning `None` **before** `owned` is constructed. Chain KKTs no longer build a per-supernode ownership map to throw away.

Two honesty notes:
- **My paired A/B is inconclusive** — 7/10, 2/10, 5/10 on my proxies. They have far fewer supernodes than clnlbeam's 39,582 and the effect sits under this container's noise floor. The 1.01–1.03× is your number, not mine; I've recorded it as such in the commit rather than claiming a win I couldn't reproduce.
- **One detail in your diagnosis is slightly off**, and it's worth correcting for the record: `vec[Vec::new(); n_snodes]` does *not* heap-allocate 39.6k Vecs — `Vec::new()` is non-allocating until first push. The real cost is the ~950 KB outer `Vec` plus the single owner vec growing to `n_snodes` entries (all inline nodes land in one task's list when `seeds=1`). Waste is real, mechanism is a bit different, fix is the same.

### 3. `bench_schur_micro` — you're right, and it's worse than "currently doesn't"

Its PACKED arm is a **local reimplementation** (`pack_a` / `run_packed` defined in the example); only the BASELINE arm calls into `feral::dense::schur_kernel`. It was written during issue #99 to justify the packed design *before it existed in the library*, and was never wired to the real thing. So `FERAL_PACKED_SIMD` correctly does nothing to it, and it was never going to move for this PR.

That's a genuine trap — it cost you time and it would cost the next person the same. The header now says so explicitly and points at `bench_dense_front` for the shipped path and `packed_matches_scalar_reference_bit_for_bit` for the byte-exactness gate. Re-pointing it at the library kernel is worth doing, but it's a bench change with no correctness content, so I've left it out of the release.

### 4. Your retraction

Thank you for measuring it and posting the reversal — that's the second time today the paired-measurement discipline overturned a plausible story, and it means the strategy is now on firm ground: **the chain gap is per-front fixed cost, not lane width**, so scaling warm-start and `nemin` are the real next round, exactly where your ranking puts them.

Your §5 sharpens that further, and I've taken the point: `nemin` is a **clnlbeam/steering fix** (36% / 15% of loop) and worth ~nothing on rocket, dtoc1nd or dtoc2, whose loops live in the 33–64 and 65–128 buckets. "Chain KKT is many small fronts" is true of some chain KKTs, not the class — same shape as your `dtoc1nd` counterexample. That belongs in the plan, and I'd rather it be scoped that way up front than discovered mid-sweep.

### Where the NEON result leaves my priors

I expected the packed kernel might be a wash on aarch64 since NEON is baseline and LLVM had a fair chance to vectorize the old tile walk. It didn't: 3.19× at n=2955, and on 0.14.0 the FMA path was buying *nothing* (11.3 vs 10.8 GF/s) where it now gets 34.1 vs 17.2. The scalar tile-walk was never an x86-only story — I'd under-called it, and the escape hatch turns out to be unnecessary rather than load-bearing.

### Release

Everything on the checklist is now green from your side: correctness gate PASS, corpus 156,929 matrices at 100% inertia on both paths, 4/4 partitions, worst residuals reproducing *to the digit* across architectures. Nothing in this commit changes numerics — 84/84 green including the `FERAL_PAR_MIN_SEEDS` parity sweep — so it doesn't disturb any of that.

I've deliberately not touched the version bump or tag: that's a call for the maintainer, not something I should do unprompted on a review branch. Say the word and I'll prepare the 0.15.0 bump (`Cargo.toml`, `python/Cargo.toml`, CHANGELOG section move) as a separate commit.

---

Issue #153, opened 2026-08-09
[perf] dtoc1nd: 3.8x MA57, concentrated in 148 fronts of ncol~62; not a dispatch gate


jkitchin:
Downstream follow-up to #148 / #150, now that 0.15.0 is released and pinned in
pounce ([jkitchin/pounce#566](https://github.com/jkitchin/pounce/pull/566)).

0.15.0 delivered downstream. Re-measuring the FERAL-vs-MA57 comparison that
motivated #148 — factorization-only through pounce's
`SparseSymLinearSolverInterface`, medians of 15, refinement off, MA57 pinned
single-threaded, FERAL left rayon-parallel, Apple M-series:

| matrix | 0.14.0 | 0.15.0 | gain | MA57 | ratio was | ratio now |
|---|---:|---:|---:|---:|---:|---:|
| clnlbeam | 57.80 ms | 25.74 | **2.25×** | 7.27 | 8.05× | **3.54×** |
| steering_12800 | 47.34 | 27.44 | **1.73×** | 18.96 | 2.53× | **1.45×** |
| rocket_12800 | 24.70 | 19.11 | 1.29× | 8.79 | 2.87× | 2.17× |
| dtoc2 | 139.84 | 109.01 | 1.28× | 84.21 | 1.70× | 1.29× |
| dtoc1nd | 16.84 | 13.67 | 1.23× | 3.63 | 4.85× | 3.77× |
| marine_1600 | 39.38 | 35.50 | 1.11× | 13.50 | 3.05× | 2.63× |

`number_of_neg_evals` is identical to 0.14.0 and to MA57 on all six.

## The reason I'm opening this: the next round's premise moved

The closing exchange on #148 scoped the next round as warm-path scaling plus
`nemin` amalgamation, with `nemin` correctly identified as a
**clnlbeam/steering fix** worth ~nothing on rocket, dtoc1nd or dtoc2. That
scoping was reasoned from clnlbeam being the 8× leader.

It is no longer the leader. **`dtoc1nd` is now the worst case at 3.77×**, and
it is the one matrix in the set that `nemin` was explicitly scoped as not
helping — nnz/dim 23.0, fronts of 33–64 columns, only 25 of its 3,831
supernodes in the `<=8` bucket (0.1% of loop time). So the highest-ratio
matrix and the planned optimization now point in different directions.

The ratio ordering also no longer tracks sparsity, which was the heuristic
behind "chain KKT is many small fronts":

[...]

PR #152, opened 2026-08-09
docs: session checkpoint 2026-08-09-04 (ship 0.15.0)

jkitchin:
Session checkpoint for the release session, plus `dev/plans/release-0.15.0-checklist.md` §3 checked off in full. Docs only, no source changes.

## What shipped

PR #151 merged as `808babb`. Main CI went green on the squash commit before tagging (different SHA from the PR head, so it needed its own run). `v0.15.0` tagged at `808babb`, GitHub Release published, both gated workflows passed:

- **Release job** — tag/`Cargo.toml` version check PASS, `cargo test` PASS, `cargo publish` PASS for all seven crates in dependency order.
- **Python wheels job** — sdist + four wheels, PyPI publish 23s, `uv pip` smoke test PASS.

Verified against the registries rather than the workflow exit codes: crates.io `feral` max_version **0.15.0**; PyPI **`feral-solver` 0.15.0** with macosx universal2, manylinux_2_17_x86_64, manylinux_2_28_aarch64, win_amd64 and the sdist.

pounce is notified ([pounce#552 comment 5232312068](https://github.com/jkitchin/pounce/issues/552#issuecomment-5232312068)) with the pickup instruction: bump the pin at `Cargo.toml:127` from `feral = { version = "0.14.0" }` to `"0.15.0"` and drop any uncommitted `[patch.crates-io]` redirect.

**Checklist §3 is complete. 0.15.0 is shipped and announced.**

## No bench run

Deliberate, and stated as such in the checkpoint: `808babb` differs from `fad5670` only in six version strings and a CHANGELOG heading, so a corpus run would re-measure identical code. The numbers stand from session 03 and the aarch64 revalidation at `6fd12d4` (156,929 matrices, dense and sparse inertia both 100.0%, 4/4 Phase 2.8.1 partitions PASS). `cargo test` for this exact SHA ran green inside the release job.

## Gotcha recorded

The PyPI distribution is named **`feral-solver`**, not `feral`. `pypi.org/pypi/feral/json` is an unrelated third-party package currently at 1.0.0, so a naive post-release check misreads the version. Any verification script must use `feral-solver`.

## Next session

Item 1 is now the measurement the release was for: re-run the pounce#552 headline models against released 0.15.0, chain-structured models first. The 3.5 to 4.8x factorization gap in that report describes the pre-SIMD kernel.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #125, opened 2026-07-10
[perf] Assembly row_indices and child→parent maps are rebuilt every numeric factorization — precompute the static structure at analysis time (SSIDS-style), splice delayed columns numerically


jkitchin:
Found in the 2026-07-10 full audit (refs are `main` @ `660224d`). Companion to the prologue issue (#124); together they cover the repeated-factorization overhead vs SSIDS.

## Defect

Every numeric factorization rebuilds, per front: `row_indices` via `build_row_indices` (`src/numeric/factorize.rs:2368`, `:3853`) — re-scanning the symmetrized A pattern plus every child's trailing rows and re-sorting (`:3910-3935`) — and the global row→local `row_map` (`:2425-2427`, populated and cleared per front). SSIDS precomputes the per-node `rlist`/`cmap` at analysis and reuses them across all numeric factorizations of the same pattern; numeric assembly is a pure scatter.

The genuinely dynamic part — **which columns are delayed** — is value-dependent and cannot be precomputed by either solver. The recoverable cost is the *static* trailing-row structure (A-pattern rows + child trailing rows), which is a pure function of the symbolic structure. The `SmallLeafSubtree` case already caches exactly this (`src/symbolic/small_leaf.rs:49,80`); the gap is interior nodes.

## How to verify the problem

Profile a repeated-factorization loop (same pattern, new values, ≥20 iterations) on a corpus KKT with deep trees: `build_row_indices` + map population recur identically every iteration. (After #124 lands, this becomes the next-largest repeated cost.)

## Proposed implementation sketch

1. At analysis: for each supernode, store the static trailing row list (A-pattern ∪ children trailing rows, sorted) and per-child scatter maps (child contrib position → parent position) computed against that static list.
2. At numeric time: delayed columns from children prepend/insert into the fully-summed region exactly as today (`factorize.rs:2332-2338`); the static maps cover the non-delayed remainder, and only the delayed splice is computed dynamically (it is small: `n_delayed` per child).
3. Memory: this stores O(Σ front heights) extra indices — bound it and report in the memory plan (`peak_contrib_bytes` accounting), with an opt-out if it matters for the largest problems.

## Validation of the fix

- Bit-identical factors/inertia on the full corpus vs `main` (assembly order must be preserved — the bit-exactness contract for the parallel driver depends on child order; keep it).
- Repeated-factorization benchmark: measurable drop in numeric-phase wall on tree-heavy KKTs; record before/after per session protocol.
[...]

PR #157, opened 2026-08-09
fix(mc64): scaling-cache gate is a drift measure + bit-identical Hungarian heap speedup (MA57 headline retracted)

jkitchin:
Fourteen commits across four sessions. The branch started as the 0.15.0 docs
push plus a chain-KKT measurement against MA57; that measurement turned out to
be wrong and retracting it is what pointed at MC64, which is where most of the
work ended up. The original write-up is kept verbatim at the bottom — it is the
record of the retraction, not stale text.

## What's in it

**Code — MC64 scaling cache gate (`bd1bc26`).** The value-bound gate deciding
whether a cached MC64 scaling may be reused had three conditions; two measured
drift from the matrix the scaling was computed on, the third did not — it
compared the current **minimum** scaled diagonal against the baseline **mean**.
Different statistics, so it behaved as an absolute property of the matrix and
rejected reuse even when the matrix had not moved. Condition 3 is now the
disjunction of the original absolute floor and a drift bound against the
baseline minimum: a strict widening, so nothing accepted today can start being
rejected. Across 53 gate evaluations on the seven corpus families that route to
MC64, exactly two decisions change — both on `robot_1600`, cutting that
trajectory's factorization time 14.3% and its scaling time 39.8%. Inertia is
unchanged on every iterate of every family, and a genuine diagonal collapse
(`arki0003`, eight orders down) is still rejected.

**Code — inline-key Hungarian heap (`509f0ce`).** The matching's binary min-heap
stored only row indices and looked the key up through a random access into the
distance array on every comparison; it now stores the key inline. Comparisons
and tie-breaking are unchanged, so the matching is unchanged.
**Verified bit-identical on 51 matrices across 39 families** by hashing the raw
bits of the scaling vector each factorization produces. Median of 3:
`nql180_0000` 1.040x, `nql180_0002` 1.057x. `pinene_3200` and `marine_1600` are
flat, as the regime table predicts — the win is in the sift-down path, which
only matters once the heap outgrows cache.

**Negative result worth the review time.** The Hungarian *search bound*, which a
prior note of mine recommended as "same-output-less-work, needs no gate and no
residual argument", does not exist as a lever. The classic
shortest-augmenting-path bound (`csp`) is already implemented
(`src/scaling/hungarian.rs:541`, `:570`, `:591`), and any further truncation
ends a search before its shortest path is proven — suboptimal matching,
different scaling vector, a numerics change needing a corpus study and human
approval. `dev/tried-and-rejected.md:2313` separately *proves* no per-column
reduced-cost bound can ever prune the inner scan and already forbids retrying
it. The correction is recorded in the new research note rather than by editing
the old one.

**Benchmark harness (`782cf65`, `197c7be`).** The MA57 oracle was timing MC64
scaling inside `MA57BD`; retracted below. Both solvers now factorize
pre-scaled matrices offline so they see identical numbers.

**Diagnostics (`16b423c`, `8f8aa8c`).** Trajectory-level scaling profile, a
scaling-reuse safety check, and issue #125 step 2 sized before building it —
the sizing said not to build it.

## Verification

- Remote CI green on `5cc64a8`: `check` x2, `stress-smoke` x2, wheels
  3.10/3.12/3.13.
- `cargo test --release`: 807 passed, 0 failed.
- Scaling-vector bit fingerprint: 51/51 matrices match baseline, re-confirmed
  after the `origin/main` merge.

## Known caveat, not attributed

The `factor/MUMPS` tail is slightly worse than the last full-corpus bench
(`2026-08-09-09`): geomean 0.43 -> 0.44, p90 1.54 -> 1.57, p99 3.30 -> 3.45,
max 8.70 -> 12.40. All Phase 2.8.1 exit partitions still PASS. It is **not** the
MC64 change — that is bit-identical, and the heap does not run on the n=225-458
CUTEst matrices (`KIRBY2`, `GROUPING`) that set those tails. Main's
`Supernode.nrow` fix, flagged in its own checkpoint as a parallel-dispatch
behavior change, is the one plausible candidate in the diff but is unverified.
A single-matrix `max` swing is also within this harness's run-to-run noise.
Recorded in `dev/sessions/2026-08-10-02.md`; next session re-benches
`origin/main` alone to separate the two.

---

# Original PR description (the MA57 retraction)

> **CORRECTION (supersedes the MA57 table below).** The "no gap" result
> was produced with a misconfigured oracle and is wrong. `ma57_bench.F`
> sets `ICNTL(15) = 1`, which makes MA57 compute MC64 scaling *inside*
> `MA57BD` — the region the oracle times. feral computes its MC64 in
> **analysis**, which `factor_us` excludes. The comparison charged MA57
> per-factorization for work feral had already amortized and was not
> being billed for.
>
> Re-run with `ICNTL(15) = 0`, 15 pairs, `min factor_us`:
>
> | matrix | feral | ma57 | MA57 faster by | wins | p |
> |---|---|---|---|---|---|
> | dtoc1nd | 14,050 | 2,539 | 5.53x | 0/15 | 0.0001 |
> | clnlbeam | 21,979 | 5,384 | 4.08x | 1/15 | 0.0010 |
> | rocket_12800 | 22,266 | 5,955 | 3.74x | 0/15 | 0.0001 |
> | marine_1600 | 33,131 | 10,973 | 3.02x | 0/15 | 0.0001 |
> | steering_12800 | 34,517 | 17,097 | 2.02x | 0/15 | 0.0001 |
> | dtoc2 | 81,725 | 51,309 | 1.59x | 0/15 | 0.0001 |
>
> MA57 is faster on **all six**. Three independent checks say this is
> the real number:
>
> - #153 measured the same six through pounce and got 1.29x-3.77x in
>   MA57's favor. Its MA57 times (clnlbeam 7.27 ms, rocket 8.79,
>   steering 18.96) match `ICNTL(15)=0` (5.38, 5.96, 17.1), not the
>   oracle's (201, 554, 292).
> - The proxy conclusion this PR claimed to supersede — 1.4x-2.4x
>   slower — was right in direction and close in magnitude. A correct
>   finding was overturned by a worse measurement.
> - The steering accuracy caveat was the same artifact. With scaling
>   off MA57 gets `1.59e-14`, not `4.53e-08`.
>
> Results 2 and 3 are unaffected: they compare feral builds to each
> other and never touch the oracle. The #150 attribution, the dtoc1nd
> regression, and the refuted E-core hypothesis all stand.
>
> What the checking missed: `factor_us` was verified to cover `MA57BD`
> only, and that was treated as sufficient. The boundary of the timed
> region was checked; the contents were not.
>
> **Confirmed by a clean experiment** ([comment](https://github.com/jkitchin/feral/pull/157#issuecomment-5234058841)).
> The table above still had feral doing MC64 in analysis and MA57 doing
> none. Pre-scaling both arms offline (`prescale_mtx` writes `D A D`;
> feral `Identity`, MA57 `ICNTL(15) = 0`) so they factorize identical
> numbers gives the same answer: MA57 faster by 4.72x / 3.56x / 2.92x /
> 2.64x / 1.46x, all 0/15 at p = 0.0001. `steering_12800` is excluded
> there — MA57's residual is `2.12e-05` against feral's `1.62e-14` on
> the same pre-scaled input, so it cannot carry a timing.

---

Supersedes #152. Docs, research notes and benchmark harness. **No solver source changed.**

## What changed since the first version of this PR

The original PR shipped a proxy measurement and asked for three
follow-ups on a machine with `data/matrices/`. Those have now been run
on an M4 Pro (`Mac16,11`, 10P+4E), and **both of the proxy's headline
conclusions are contradicted.** The proxy note is kept, marked
superseded, with the contradiction stated at the top — it is a useful
record of geometry-matched stand-ins giving the wrong answer.

New: `dev/research/chain-kkt-corpus-2026-08-09.md`.

## Result 1 — feral is not behind MA57 on chain KKTs

Six real KKTs from `data/matrices/kkt-mittelmann`, paired alternating
A/B, `min factor_us` over 15 pairs, exact sign test. Ratio > 1 means
feral is faster.

| matrix | n | main | ma57 | ratio | wins | p |
|---|---|---|---|---|---|---|
| rocket_12800 | 89,601 | 22,893 | 535,659 | **23.40** | 15/15 | 0.0001 |
| clnlbeam | 99,999 | 21,739 | 210,481 | **9.68** | 15/15 | 0.0001 |
| steering_12800 | 115,201 | 33,923 | 273,494 | **8.06** | 15/15 | 0.0001 |
| marine_1600 | 76,807 | 33,373 | 42,471 | 1.27 | 15/15 | 0.0001 |
| dtoc2 | 103,920 | 81,163 | 61,245 | 0.76 | 0/15 | 0.0001 |
| dtoc1nd | 9,685 | 14,297 | 4,561 | 0.32 | 0/15 | 0.0001 |

The proxies said "slower than MA57 on all five, 1.4x to 2.4x, worst on
the largest and most chain-like." On the real matrices the largest and
most chain-like are where feral wins hardest.

**Caveat that belongs with the 8.06x:** on `steering_12800` MA57's
residual is `4.53e-08` against feral's `3.52e-14`, same matrix and RHS.
feral reports `refined yes`, so part of that is iterative refinement
the MA57 oracle does not do. `factor_us` is factorization only so the
timing stands, but feral is faster there *and* more accurate, against a
looser MA57 solve. The correctness gate in the new runner is what
surfaced this; a timing-only table would have hidden it.

## Result 2 — the regression is real, and it is on the smallest matrix

Ratios vs `v0.14.0`. All large-matrix gains and the `dtoc1nd` loss are
15/15 at p = 0.0001.

| matrix | #149 SIMD | #150 parallel | v0.15.0 | main |
|---|---|---|---|---|
| clnlbeam | 0.966 | 2.048 | 2.054 | 2.048 |
| steering_12800 | 0.986 | 1.646 | 1.644 | 1.669 |
| dtoc2 | 0.999 | 1.401 | 1.438 | 1.434 |
| rocket_12800 | 0.979 | 1.200 | 1.255 | 1.214 |
| marine_1600 | 0.953 | 0.956 | 0.953 | 0.943 |
| **dtoc1nd** | 0.960 | **0.764** | **0.762** | **0.750** |

#150 (`fad5670`) is the mover and it is a large win on the four
largest. The one regression is `dtoc1nd`, n = 9,685 — 25%, introduced
by #150, carried unchanged into v0.15.0 and main. The proxies got the
existence of a regression right and its location exactly backwards.

v0.15.0 and main are indistinguishable, so the tag-vs-post-tag question
this PR flagged as open is closed: neither #155 nor #156 moves anything
here.

#149's SIMD kernel is below 1.0 on all six, significantly on `clnlbeam`
(0.966, 0/15, p = 0.0001). It helped on the proxies. That deserves its
own bisect.

## Result 3 — thread count does nothing; the E-core hypothesis is refuted

`RAYON_NUM_THREADS` = 1/2/4/8/10 vs default: every ratio within 6% of
1.0, one significant and in the wrong direction. On a 10P+4E machine,
where the predicted effect should be larger than on the 4P+4E M2 that
generated the hypothesis. Recorded in `tried-and-rejected.md`.

The knob was verified live first — feral reads the global rayon pool
(`rayon::current_num_threads()`), and the same variable moved the
proxies by up to 65%.

**Consequence:** single-threaded main matches all-threads main on every
matrix, so **#150's gains are not parallelism gains.** Any follow-up
that tries to extend #150 by parallelizing harder starts from a false
premise.

## steering_12800 was missing; the harvest path is dead

The corpus shipped only a zero-byte `.solver.log` for it. Cause: ripopt
commit `76d3575` deleted the KKT dump implementation while leaving
`SolverOptions::kkt_dump_dir` and its CLI wiring in place, so
`kkt_dump_dir=` still parses and silently writes nothing — the harvest
reports "OK ... 0 mtx" and exits 0.

Regenerated with pounce (`--dump kkt:1-3`) plus a new
`scripts/harvest-pounce-kkt.py`. Validated against an external oracle
rather than itself: pounce reports `num_neg_evals_actual = 51201`, and
feral and MA57 independently agree (`inertia_neg = 51201`,
`inertia_zero = 0`). This also unblocks the 41 Mittelmann families that
were never harvested.

`dtoc2_0001`/`_0002` are excluded: singular by both solvers
(`inertia_zero` 4 and 103, both `rel_res = NaN`). `dtoc2_0000` is used
instead.

## Harness

- `arm_run.py` — `ab_run.py` generalized to N arms with per-arm env
  overrides, reusing its protocol functions so the statistics are the
  same code. This is the `probe_bisect.py` the original note recorded
  as written-but-never-run. Prints a correctness gate before any timing
  table.
- `real_corpus_mtx.py` — builds the matrix set from
  `data/matrices/kkt-mittelmann`, selection rules in the docstring.
- `scripts/harvest-pounce-kkt.py` — pounce JSONL dump to corpus
  `.mtx`/`.json`.

## Limits

One iterate per family, six families, one machine, aarch64 only.
`factor_us` is numeric factorization only, and analysis dominates wall
clock on these matrices (`clnlbeam` analyse 4.59 s vs factor 38 ms), so
this still does not directly answer pounce#552's end-to-end numbers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed — MC64 scaling cache rejected matrices against their own fingerprint
+
+- The value-bound gate that decides whether a cached MC64 scaling may be reused
+  had three conditions, two of which measure how far the matrix has drifted from
+  the one the scaling was computed on. The third did not: it compared the
+  current **minimum** scaled diagonal against the baseline **mean**. Because
+  those are different statistics, it behaved as an absolute property of the
+  matrix rather than a drift measure, and rejected reuse even when the matrix
+  had not moved at all.
+- Condition 3 is now the disjunction of the original absolute floor and a drift
+  bound against the baseline minimum. This is a strict widening: nothing the
+  gate accepts today can start being rejected.
+- Effect is narrow and measured. Across 53 gate evaluations on the seven corpus
+  families that route to MC64, exactly two decisions change — both on
+  `robot_1600`, cutting that trajectory's factorization time 14.3% and its
+  scaling time 39.8%. Inertia is unchanged on every iterate of every family, and
+  a genuine diagonal collapse (`arki0003`, eight orders down) is still rejected.
+- No API change; the gate is internal.
+
 ### Changed — `Solver` defaults `use_parallel` from the platform (issue #154)
 
 - `Solver::new()` and `Solver::with_params(...)` now derive `use_parallel` from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Changed — MC64 Hungarian matching is 4-5% faster on large matchings *(bit-identical)*
+
+- The Hungarian matching's binary min-heap stored only row indices and looked
+  the key up through a random access into the distance array on every
+  comparison. It now stores the key inline alongside the index. Comparisons
+  and tie-breaking are unchanged, so the matching — and therefore the scaling
+  vector — is unchanged.
+- **Verified bit-identical on 51 matrices across 39 families**, by hashing the
+  raw bits of the scaling vector each factorization produces.
+- Measured (median of 3): `nql180_0000` 1.040x, `nql180_0002` 1.057x. Matrices
+  whose matchings are small and L1-resident (`pinene_3200`, `marine_1600`) are
+  unchanged, as expected — the win is in the sift-down path, which only
+  matters when the heap outgrows cache.
+- No API change.
+
 ### Fixed — MC64 scaling cache rejected matrices against their own fingerprint
 
 - The value-bound gate that decides whether a cached MC64 scaling may be reused
--- a/crates/feral-diagnostics/src/bin/bench_one_matrix.rs
+++ b/crates/feral-diagnostics/src/bin/bench_one_matrix.rs
@@ -21,7 +21,7 @@ use std::time::Instant;
 use feral::numeric::factorize::factorize_multifrontal_parallel_with_workspace;
 use feral::numeric::factorize::FactorWorkspace;
 use feral::numeric::solve::solve_sparse_refined;
-use feral::scaling::{Mc64FallbackReason, ScalingInfo};
+use feral::scaling::{Mc64FallbackReason, ScalingInfo, ScalingStrategy};
 use feral::symbolic::{symbolic_factorize, SupernodeParams};
 use feral::{read_mtx, CscMatrix, NumericParams};
 
@@ -141,7 +141,15 @@ fn solve_one(mtx_path: &str, rhs_path: &str, out_path: &str) -> std::io::Result<
     // InfNorm picked by matrix shape) and BK pivot threshold = 1e-8.
     // This matches the high-level `Solver::new()` defaults; bench
     // measures what library users get out of the box.
-    let params = NumericParams::default();
+    let mut params = NumericParams::default();
+    // `FERAL_SCALING=identity` disables feral's own scaling so the
+    // factorization can be timed against a solver that is also doing
+    // none -- see `prescale_mtx`, which bakes `D A D` into the matrix
+    // so both arms factorize identical numbers. Anything else (or
+    // unset) leaves the out-of-the-box `Auto` default alone.
+    if env::var("FERAL_SCALING").as_deref() == Ok("identity") {
+        params.scaling = ScalingStrategy::Identity;
+    }
     let mut ws = FactorWorkspace::new();
     let t0 = Instant::now();
     let factor_res = factorize_multifrontal_parallel_with_workspace(&csc, &sym, &params, &mut ws);
@@ -188,6 +196,7 @@ fn solve_one(mtx_path: &str, rhs_path: &str, out_path: &str) -> std::io::Result<
     writeln!(out, "inertia_neg {}", inertia.negative)?;
     writeln!(out, "inertia_zero {}", inertia.zero)?;
     writeln!(out, "analyse_us {}", analyse_us)?;
+    writeln!(out, "scaling {:?}", params.scaling)?;
     writeln!(out, "factor_us {}", factor_us)?;
     writeln!(out, "solve_us {}", solve_us)?;
     writeln!(out, "rel_res {:.17e}", rel)?;
--- a/crates/feral-diagnostics/src/bin/diag_delayed_fronts.rs
+++ b/crates/feral-diagnostics/src/bin/diag_delayed_fronts.rs
@@ -1,0 +1,112 @@
+//! diag_delayed_fronts — how often does issue #125's static-layout fast
+//! path miss?
+//!
+//! Step 1 of #125 (PR #139) reuses the symbolic static frontal layout
+//! only when `n_delayed_in == 0`. Step 2 (still open) would extend it to
+//! fronts that receive delayed columns. This reports what fraction of
+//! fronts, and what fraction of frontal *rows*, take the dynamic
+//! `build_row_indices` path today — i.e. the ceiling on what step 2 can
+//! recover.
+//!
+//! Usage: diag_delayed_fronts <a.mtx> [b.mtx ...]
+
+use feral::dense::factor::{phase_timing, PHASE_TIMING_ENABLED};
+use feral::read_mtx;
+use feral::symbolic::SupernodeParams;
+use feral::{NumericParams, Solver};
+use std::path::Path;
+use std::sync::atomic::Ordering::Relaxed;
+use std::time::Instant;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args: Vec<String> = std::env::args().skip(1).collect();
+    if args.is_empty() {
+        eprintln!("usage: diag_delayed_fronts <a.mtx> [b.mtx ...]");
+        std::process::exit(2);
+    }
+    println!(
+        "{:<24}{:>9}{:>10}{:>9}{:>11}{:>12}{:>11}",
+        "matrix", "fronts", "delayed", "pct", "rows", "rows_dyn", "pct_rows"
+    );
+    let mut buildrow_rows = Vec::new();
+    for a in &args {
+        let csc = match read_mtx(Path::new(a)).and_then(|m| m.to_csc()) {
+            Ok(c) => c,
+            Err(e) => {
+                eprintln!("skip {a}: {e:?}");
+                continue;
+            }
+        };
+        let mut solver = Solver::with_params(NumericParams::default(), SupernodeParams::default())
+            .with_parallel(false);
+        let status = solver.factor(&csc, None);
+        if !matches!(status, feral::FactorStatus::Success) {
+            eprintln!("{a}: factor status {status:?}");
+        }
+        let f = match solver.factors() {
+            Some(f) => f,
+            None => {
+                eprintln!("{a}: no factors");
+                continue;
+            }
+        };
+        let (mut fronts, mut delayed, mut rows, mut rows_dyn) = (0usize, 0usize, 0usize, 0usize);
+        for nf in &f.node_factors {
+            fronts += 1;
+            rows += nf.nrow;
+            if nf.n_delayed_in > 0 {
+                delayed += 1;
+                rows_dyn += nf.nrow;
+            }
+        }
+        // Warm refactor with phase timing on: `BUILDROW_NS` wraps both
+        // arms of the #125 branch (static `.to_vec()` copy and dynamic
+        // `build_row_indices`), so buildrow/factor is a hard ceiling on
+        // what step 2 could ever recover.
+        PHASE_TIMING_ENABLED.store(true, Relaxed);
+        phase_timing::reset();
+        let t0 = Instant::now();
+        let _ = solver.factor(&csc, None);
+        let factor_ns = t0.elapsed().as_nanos() as u64;
+        PHASE_TIMING_ENABLED.store(false, Relaxed);
+        let (buildrow_ns, ..) = phase_timing::snapshot_detail();
+
+        let name = Path::new(a)
+            .file_stem()
+            .and_then(|s| s.to_str())
+            .unwrap_or(a);
+        println!(
+            "{:<24}{:>9}{:>10}{:>8.2}%{:>11}{:>12}{:>10.2}%",
+            name,
+            fronts,
+            delayed,
+            100.0 * delayed as f64 / fronts.max(1) as f64,
+            rows,
+            rows_dyn,
+            100.0 * rows_dyn as f64 / rows.max(1) as f64
+        );
+        buildrow_rows.push((
+            name.to_string(),
+            factor_ns,
+            buildrow_ns,
+            100.0 * rows_dyn as f64 / rows.max(1) as f64,
+        ));
+    }
+
+    println!();
+    println!(
+        "{:<24}{:>13}{:>13}{:>12}{:>12}",
+        "matrix", "factor_us", "buildrow_us", "buildrow%", "dyn_rows%"
+    );
+    for (name, factor_ns, buildrow_ns, pct_rows) in &buildrow_rows {
+        println!(
+            "{:<24}{:>13}{:>13}{:>11.2}%{:>11.2}%",
+            name,
+            factor_ns / 1000,
+            buildrow_ns / 1000,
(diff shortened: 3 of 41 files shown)


jkitchin:
> **CORRECTION (supersedes the MA57 table below).** The "no gap" result
> was produced with a misconfigured oracle and is wrong. `ma57_bench.F`
> sets `ICNTL(15) = 1`, which makes MA57 compute MC64 scaling *inside*
> `MA57BD` — the region the oracle times. feral computes its MC64 in
> **analysis**, which `factor_us` excludes. The comparison charged MA57
> per-factorization for work feral had already amortized and was not
> being billed for.
>
> Re-run with `ICNTL(15) = 0`, 15 pairs, `min factor_us`:
>
> | matrix | feral | ma57 | MA57 faster by | wins | p |
> |---|---|---|---|---|---|
> | dtoc1nd | 14,050 | 2,539 | 5.53x | 0/15 | 0.0001 |
> | clnlbeam | 21,979 | 5,384 | 4.08x | 1/15 | 0.0010 |
> | rocket_12800 | 22,266 | 5,955 | 3.74x | 0/15 | 0.0001 |
> | marine_1600 | 33,131 | 10,973 | 3.02x | 0/15 | 0.0001 |
> | steering_12800 | 34,517 | 17,097 | 2.02x | 0/15 | 0.0001 |
> | dtoc2 | 81,725 | 51,309 | 1.59x | 0/15 | 0.0001 |
>
> MA57 is faster on **all six**. Three independent checks say this is
> the real number:
>
> - #153 measured the same six through pounce and got 1.29x-3.77x in
>   MA57's favor. Its MA57 times (clnlbeam 7.27 ms, rocket 8.79,
>   steering 18.96) match `ICNTL(15)=0` (5.38, 5.96, 17.1), not the
>   oracle's (201, 554, 292).
> - The proxy conclusion this PR claimed to supersede — 1.4x-2.4x
>   slower — was right in direction and close in magnitude. A correct
>   finding was overturned by a worse measurement.
> - The steering accuracy caveat was the same artifact. With scaling
>   off MA57 gets `1.59e-14`, not `4.53e-08`.
>
> Results 2 and 3 are unaffected: they compare feral builds to each
> other and never touch the oracle. The #150 attribution, the dtoc1nd
> regression, and the refuted E-core hypothesis all stand.
>
> What the checking missed: `factor_us` was verified to cover `MA57BD`
> only, and that was treated as sufficient. The boundary of the timed
> region was checked; the contents were not.

jkitchin:
## Clean experiment: both solvers on identical numbers

The correction above still compared arms doing different work — feral
was computing its own MC64 during analysis, MA57 was doing none. This
removes that asymmetry entirely.

`prescale_mtx` computes MC64 symmetric scaling once and writes
`A' = D A D` to disk. Both solvers read that file. feral runs
`FERAL_SCALING=identity` (`ScalingStrategy::Identity`), MA57 runs
`ICNTL(15) = 0`. Neither scales; `factor_us` is factorization and
nothing else. All six pre-scaled with a real matching
(`scaling_info=Applied`, no infinity-norm fallback). Both arms echo
their strategy into the sidecar, so the configuration is checkable
after the fact: `scaling Identity` / `scaling 0`.

15 pairs, `min factor_us`, exact two-sided sign test.

| matrix | feral | ma57 | MA57 faster by | wins | p |
|---|---|---|---|---|---|
| dtoc1nd | 12,061 | 2,555 | 4.72x | 0/15 | 0.0001 |
| clnlbeam | 19,315 | 5,418 | 3.56x | 0/15 | 0.0001 |
| marine_1600 | 28,949 | 9,924 | 2.92x | 0/15 | 0.0001 |
| rocket_12800 | 16,332 | 6,184 | 2.64x | 0/15 | 0.0001 |
| dtoc2 | 75,291 | 51,757 | 1.46x | 0/15 | 0.0001 |

`steering_12800` is **excluded, not quoted**. MA57's residual on the
pre-scaled matrix is `2.12e-05` against feral's `1.62e-14` on the same
input — MA57 does not solve it acceptably without its own scaling, so
that row cannot carry a timing. The correctness gate caught it, as it
caught the `4.53e-08` in the run before. Every other cell is `1e-16` to
`6e-14` on both arms.

Three protocols now agree — different harnesses, different scaling
handling, different machines (MA57 faster by):

| matrix | issue #153 (pounce) | ICNTL(15)=0, unscaled | clean, pre-scaled |
|---|---|---|---|
| dtoc1nd | 3.77x | 5.53x | 4.72x |
| clnlbeam | 3.54x | 4.08x | 3.56x |
| marine_1600 | 2.63x | 3.02x | 2.92x |
| rocket_12800 | 2.17x | 3.74x | 2.64x |
| dtoc2 | 1.29x | 1.59x | 1.46x |

**MA57's numeric factorization is roughly 1.5x to 4.7x faster than
feral's on these matrices.** The gap #148 and #153 have been tracking
is real and 0.15.0 did not close it. The retracted result was the only
measurement that ever said otherwise.

Secondary reading: `feral` with default `Auto` scaling ran within 2-8%
of `feral` with scaling off, on these already-scaled inputs. That is
not in tension with #153's 18-34% warm-path scaling cost — MC64 on an
already-MC64-scaled matrix converges immediately — but it does mean
this experiment says nothing about warm-path scaling, which needs an
unscaled input to measure.

jkitchin:
## Second correction, and what profiling MC64 actually found

**Result 4 had the same error as the retracted result, one stage up.**
It compared feral's `analyse_us` against MA57's `analyse_us` at
`ICNTL(15) = 0` and called feral's analysis 794x slower on clnlbeam.
feral's analysis does MC64; MA57's analysis at `ICNTL(15) = 0` does no
scaling at all.

Charging MA57 for its scaling (its `factor` at `ICNTL(15) = 1` minus at
`= 0`, both already measured):

| matrix | as reported | corrected |
|---|---|---|
| clnlbeam | 794x slower | **24.8x** |
| rocket_12800 | 421x slower | **4.9x** |
| marine_1600 | 144x slower | **26.4x** |
| dtoc1nd | 17x slower | **9.4x** |
| steering_12800 | 1.4x faster | 3.5x faster |
| dtoc2 | 1.6x faster | 1.6x faster |

MC64 head to head, both solvers doing the same job: feral 23.2x slower
(clnlbeam), 25.1x (marine), 4.2x (rocket), ~1.3x (dtoc1nd, dtoc2) —
and **13.3x faster on steering_12800**.

Calling `MA57ID` directly confirms **MA57's own default is
`ICNTL(15) = 1`**. Scaling-on is its default configuration, so
`ICNTL(15) = 0` is the right setting for a factorization-only
comparison and the wrong baseline for anything end-to-end. **The clean
experiment is unaffected** — both arms are scaling-free on pre-scaled
input — so Result 5 stands as written.

## The dense-column diagnosis does not generalize

`ldlt_compress` at 99.3% of clnlbeam's symbolic made MC64 the obvious
target, so it was profiled before any work started (`probe_mc64_hungarian`,
41 corpus matrices, one iterate per family).

`mc64-dense-column-2026-06-06.md` attributes MC64's cost to a dense
coupling column, `O(searches x dense_deg)`, and proves no
behavior-preserving inner-loop fast path exists. That is correct for
rocket_12800 and covers one of the four expensive matrices. It is not
the general law:

| matrix | `max_col_degree` | ms |
|---|---|---|
| steering_12800 | **51,202** (largest in corpus) | 21 |
| gasoil_3200 | 41,500 | 9.6 |
| rocket_12800 | 38,401 | 2,279 |
| nql180 | 182 | 2,892 |
| clnlbeam | **5** | 4,522 |

The two densest-column matrices in the corpus are among the fastest;
the most expensive one has `max_col_degree = 5`. `log(ms)` against
`log(max_col_degree)` gives r = **-0.59** — the wrong sign. Against
`touched_total`, r = **+0.999**.

## The measured law

Leading principal submatrices of clnlbeam, with `searches` pinned at
5,043 across the range:

| n | touched | edge_scans | ms |
|---|---|---|---|
| 90,000 | 176,356,074 | 360,898,556 | 4,080 |
| 96,000 | 200,410,074 | 360,892,553 | 4,351 |
| 99,999 | 216,406,073 | 360,922,194 | 4,525 |

`edge_scans` is **flat** — 3.609e8, varying under 0.01% while `n` grows
11%. `touched` grows exactly **+4,009 per unit of `n`**, with all four
independent deltas (Δn = 3000, 3000, 2000, 1000) giving 4009.0.

Each augmenting search reaches **Θ(n) distinct rows while doing a
constant amount of edge work.** Growing the problem buys pure heap
traffic and no additional edge scanning. `touched` counts distinct rows
(`hungarian.rs:549` guards the push on `d[i] == RINF`), and
`heap_init_slots == touched_total + n` holds exactly on every matrix —
so issue #80's incremental-reset fix is intact and heap *initialization*
is not the cost.

Value-driven, not structural: synthetic chain KKTs at the same geometry
up to n = 384,000 get `searches = 0` and run in 18 ms. One more case of
the proxies failing to reproduce the real matrices.

**No fix is claimed.** The impossibility proof covers the dense-column
inner scan, not this Θ(n) frontier. Whether the frontier can be bounded
without changing the matching needs its own research note before any
code.

Issue #161, opened 2026-08-13
LU: numeric kernel is 1.6-3.9x SuperLU on real LP bases at equal fill; triangular solves are not work-proportional


jkitchin:
## Summary

On real LP simplex bases, feral's sparse LU is **1.6x–3.9x slower than SuperLU at
equal or better fill**, and its triangular solves cost the same whether the
solution has 1 nonzero or `m`. The ordering is not the problem — feral's fill is
as good as or better than COLAMD's on the same matrices. The cost is in the
numeric kernel and in the solve's lack of work-proportionality.

This is the dominant cost in a downstream dual simplex (discopt#1008): on the LP
where that solver is 17.2x slower than HiGHS, **93.1% of wall is in this LU
layer**.

## Reproducer

Self-contained, no discopt dependency. Two real bases extracted from a running
dual simplex are attached as `.npz`; with no argument the script generates a
synthetic basis with the same structural signature.

```
python feral_repro.py QPLIB_1157_basis.npz     # high-fill basis
python feral_repro.py QPLIB_3852_basis.npz     # low-fill basis
python feral_repro.py                          # synthetic, no data files needed
```

Measured with feral 0.13.0 (the version in the Python wheel on this machine);
the Rust-side numbers below are feral 0.15.1, and both show the effect.

The script and both `.npz` bases are in the discopt tree at
`scratchpad/issue1008/` (GitHub will not take `.npz` as issue attachments). The
no-argument synthetic path needs no data files and reproduces the effect on its
own — more strongly than the real bases do (28.1x vs SuperLU), so treat the
synthetic number as an illustration and the real-basis numbers as the evidence.

## A. Numeric factorization: 1.6x–3.9x SuperLU at equal fill

`QPLIB_1157_basis.npz` — m=3937, nnz=29376 (7.46/col):

```
  feral      46.99 +- 0.90 ms   factor_nnz=190654   fill = 6.49x   246 ns per factor-nnz
  SuperLU    12.10 +- 0.21 ms   factor_nnz=213132   fill = 7.26x
  --> feral is 3.9x SuperLU's time at 0.89x its fill
```

`QPLIB_3852_basis.npz` — m=1760, nnz=4003 (2.27/col):

```
  feral       0.55 +- 0.08 ms   factor_nnz=4308     fill = 1.08x   127 ns per factor-nnz
  SuperLU     0.34 +- 0.10 ms   factor_nnz=4295     fill = 1.07x
[...]

PR #162, opened 2026-08-13
feat(lu): work-proportional sparse LU solves (#161B), plus the #163 ordering revert

jkitchin:
> **Updated after @jkitchin's transfer test on the real bases.** Two things
> changed: the default cap was wrong and is now `0.10`, and the end-to-end
> framing this PR was written against has been retracted. Both are folded into
> the sections below rather than appended, so this body reads as the current
> state. See the review comment for the measurements.

## What

Issue #161 part B — "triangular solves are not work-proportional". Filed as
*described, not implemented*: the issue's numbers size the prize but do not
demonstrate a fix. Part A landed in #160; this is the other half, plus the two
loose ends #160's own review left open.

**Read the "What this is worth end to end" section before sizing anything off
the component ratios below.** They are large and they are real, but on the LP
that motivated issue #161 this change is ~1.00x.

## The diagnostic was in the issue's own number

Issue #161 measured a one-nonzero `ftran` at **0.74x** the cost of a fully dense
one — "2918x more work than it performs". The 0.74x rather than 1.00x is the
tell. The four sparse triangular kernels split into two shapes:

| kernel | form | zero handling | cost |
|---|---|---|---|
| `lsolve` (`L y = s`) | scatter | skips zeros before touching `L(:,k)` | `O(m + flops)` |
| `ut_solve` (`Uᵀ z = s`) | scatter | skips — but only *after* `row.first()` | `O(m` row touches` + flops)` |
| `usolve` (`U w = s`) | **gather** | none — every row of `U` is read | `O(nnz(U))` always |
| `lt_solve` (`Lᵀ v = s`) | **gather** | none — every column of `L` is read | `O(nnz(L))` always |

Half the solve already went to ~0 and half stayed at full cost, which is exactly
what predicts a ratio between 0.5x and 1.0x. So the fix is to the two gather
kernels specifically, and the prediction is that it moves the sparse-rhs case and
leaves the dense-rhs case alone.

## Two changes

**1. Reach-limited ("hyper-sparse") sweeps — existing API, no caller change.**
`usolve` and `lt_solve` compute the reach of the rhs pattern in the factor's DAG
(Hall & McKinnon 2005) and sweep only the positions the solution depends on.
`usolve` walks `u_above`, which the Forrest–Tomlin update already builds and
maintains — nothing new was needed. `lt_solve` needed a row-wise index of `L`,
built at factor time and valid for the life of the factor because the FT update
never touches the base `L`. Both sort the reach into topological order: by
decreasing `uperm` **rank** for `U` (*not* position index — `U` is upper
triangular in `uperm` order once an update has permuted it) and by decreasing
position for `Lᵀ`.

`ut_solve` separately hoists its `s[i] == 0.0` test above the `u_rows[i]`
dereference. It was already flop-proportional but paid `m` cache misses per solve
chasing one heap allocation per row.

New `LuParams::hyper_sparse_max_density` (**default `0.10`** — see below) caps
the reach as a fraction of `m`; a reach that overruns it aborts back to the dense
sweep, bounding what a sparse rhs whose solution *fills in* can cost. `0.0`
restores the previous solve exactly and allocates none of the new state.

**2. `ftran_sparse` / `btran_sparse` — a new signature, because the floor is the
signature.**

After change 1, a solve was `O(m + reach work)` and stopped there. A phase probe
on a 3.4-position reach at `m = 4000` measured 22 us spread evenly across ~6
linear passes, with `ftran_partial` — just the `P`-gather and `lsolve` — already
10.3 of them. Every one of those passes exists because
`ftran(&mut self, rhs: &mut [f64])` hands the solve a dense vector, which is the
only description of the input it has. An attempt to attack the largest-looking of
them measured as noise and was reverted (see `dev/tried-and-rejected.md`).

```rust
pub fn ftran_sparse(&mut self, rhs: &[(usize, f64)], out: &mut Vec<(usize, f64)>)
    -> Result<(), FeralError>;
pub fn btran_sparse(...) -> Result<(), FeralError>;
```

Nothing in the path is proportional to `m`: scatter `O(nnz(rhs))`, both
triangular solves `O(reach work)`, the eta replay `O(eta ops)` (proportional to
the update chain, not to `m`), gather `O(nnz(x) log nnz(x))`.

## The default cap: 0.10, after 0.25 was measured wrong

This PR originally shipped `0.25`, from a sweep on the in-tree **synthetic**
fixture that concluded "the win is flat from 0.05 to 1.00". That number was
blended over `ftran` and `btran` and reported as one ratio. It is true for
`ftran` alone. On the real QPLIB_1157 basis:

| cap | `ftran` | `btran` | route fired |
|---|---|---|---|
| 0.05 | 10.59x | 0.98x | 1195 |
| **0.10 (now shipped)** | **10.81x** | **0.97x** | **1195** |
| 0.25 (was shipped) | 11.01x | **0.69x** | 2370 |
| 1.00 | 10.75x | **0.69x** | 2560 |

`btran` falls off a cliff between 0.10 and 0.25 — exactly where the fired count
doubles. Those extra ~1175 solves are `btran` reaches sitting in the 10–25% band,
taking the route and losing **1.45x** to the `O(r log r)` sort this PR already
listed as a known limit, for no `ftran` gain.

**The fixture structurally could not show this.** `lp_basis()` produces solutions
that are either tiny (p50 = 3) or ~9% of `m` — bimodal, with nothing in the
10–25% band. QPLIB_1157's mean solution density is 14.7% of `m`, squarely inside
it; the generator was tuned to QPLIB_3852's 2.27 nnz/col, where `btran` is 1.02x
and the cap never bites. A cap can only be validated against a basis whose
solution density straddles it.

`default_density_cap_is_pinned_to_the_measured_value` now pins the value with
that reasoning, so raising it again has to come with evidence from a basis whose
solutions live in the admitted band and *win* there.

Adopted on @jkitchin's measurement, not reproduced here — the attachment did not
arrive (no link in either comment body; not in discopt's git either). Lowering
the cap is the conservative direction: it strictly reduces the population taking
the new route, so the worst case is forgoing a win rather than introducing a loss.

## Component measurements

Interleaved A/B, two factors of one basis differing only in the parameter under
test, arm order alternating rep by rep. `examples/hyper_sparse_solve.rs` exits
non-zero if a route never fires, so none of this can be a flattering measurement
of the dense path against itself.

**Reach-limited sweeps**, real bases (@jkitchin's run, cap 0.10):

```
  QPLIB_3852  m=1760  2.27 nnz/col  fill 1.01x   ftran 1.74x    btran 1.02x
  QPLIB_1157  m=3937  7.46 nnz/col  fill 6.76x   ftran 10.81x   btran 0.97x
  max |off − on| = 0.000e0 on both; factor_nnz identical; route asserted fired
```

**Sparse API**, synthetic triangular basis, `ftran` p50, median of 3 runs, with
the deterministic operation count from `last_sparse_solve_work()`:

| `m` | dense sweep | `ftran_sparse` | ratio | **work (ops)** |
|---|---|---|---|---|
| 1,000 | 9.41 us | 0.302 us | 31x | **12.4** |
| 4,000 | 48.98 us | 1.105 us | 44x | **12.3** |
| 16,000 | 244.9 us | 1.805 us | 136x | **13.2** |
| 64,000 | 1203.8 us | 2.53 us | 476x | **11.2** |

**The work column is the result.** `m` grows 64x and the operation count does not
move. The ratio climbing with `m` is just the dense sweep's `O(m)` term growing
while the sparse path's does not.

## What this is worth end to end

Measured through discopt on the root LP of QPLIB_1157, discopt held fixed and
only feral varied:

| feral arm | wall | vs crates.io |
|---|---|---|
| 0.15.1 | 6.088 s | 1.00x |
| `main` + `dense_bump_max_dim = 4096` | 3.567 s | 1.71x |
| this PR + bump + cap 0.10 | 3.607 s | 1.69x |

**~1.00x**, against 10.6x on the component. Issue #161's "93.1% of wall in the LU
layer" has been retracted by its author as a profile-attribution artifact —
triangular solves are ~1% of that solve. Three other QPLIB instances gain 4–13%.

A 10.6x speedup that moves the total by 1% is a direct measurement that the
component was 1% of the total. This work is sound and free, but **the 476x is a
component ratio and must not be read as a downstream prediction** — the
`ftran_sparse` follow-up in particular should not be sized off it. On this
evidence the LU cost that matters for that consumer is numeric factorization
(`dense_bump_max_dim`, ≥41% of that solve's wall), not the solves.

## Other things stated rather than left to be discovered

**Mean and median say different things.** On the mixed-density synthetic fixture
the sparse API is 8.7x on the *mean* and 103x on the *median*, because the
minority of solves that reach through the whole bump dominate the mean. Where the
solution is genuinely sparse the win is large; where it is not, this path is
neutral-to-slower — it sorts its reach where the dense sweep just walks the
factor in order. **Callers with dense solutions should keep using `ftran`**, and
the doc comments say so.

**The wall clock is noisy at large `m`; the work counter is not.** One `m = 64000`
run reported a btran mean of 26.3 us where repeats gave 3.0–3.4 us. That is why
the asymptotic claim is pinned by `last_sparse_solve_work()` in a test rather
than by a benchmark: a reintroduced `O(m)` term would present as a constant
factor, indistinguishable from machine noise.

## Semantics that change

The `SingularBasis` guard on `U`'s stored diagonal (the L10 hardening) is now
evaluated on the rows a solution depends on rather than on all `m` rows of every
solve — in `ut_solve` unconditionally, and in `usolve` on the reach-limited
route.

This is sound, not merely cheaper: a skipped row `k` has `s[k] == 0` and no
reached predecessor, so back substitution would assign it `0 / U[k,k]`. For a
healthy pivot that is `0`, which is what leaving it alone already gives; for a
zero pivot the row states `0 = 0`, the system is consistent and underdetermined
there, and `0` remains *a* correct solution component. No solve returns a
different answer. What narrows is the incidental *diagnostic* that the factor is
degenerate somewhere the caller never asked about — which factor and update time
still report, against the pivot tolerance.

Recorded in `dev/decisions.md`, because it narrows a guard an earlier repo review
deliberately made always-on. Both tests that pin L10 pass unchanged.
`hyper_sparse_max_density = 0.0` restores the strict behavior.

## Vacuity guards, which earned their keep immediately

Both routes are **silent fallbacks** — a solve that overruns the cap, or a bump
that fails its structural precondition, produces the same answer by the slow
path. A test that only checks "the answer is right" therefore passes whether or
not the new code ran at all.

`hyper_sparse_sweeps()` / `hyper_sparse_nodes()` / `last_sparse_solve_work()`
exist so tests can assert the route fired. On their first run, **all nine
differential tests passed on correctness while the sweep count was 0**:
`reach_cap()` inspected `self.reach` to decide whether the workspace was live,
but the solve entry points `mem::take` that workspace out of `self` for the
duration of the call, so the test always saw the empty `Default`. Without the
counter this ships as a no-op with a green suite. Same argument #160 makes for
`used_dense_bump()`.

## Composition with #160

#160's dense-bump route **splices a dense factorization into the sparse `L` and
`U`** — and the reach-limited sweeps navigate exactly those structures. A splice
producing entries `u_above` or the `L` row index did not describe would truncate
a reach and return a wrong answer with no other symptom. Neither branch could
have caught this (part A predates the reach work; part B was written against a
tree without the splice), so it is tested here:
`reach_route_composes_with_the_dense_bump_route` and
`sparse_solves_compose_with_the_dense_bump_route`, both asserting
`used_dense_bump()` alongside their own route's counter.

Also recorded rather than glossed: with the bump route on, `nnz(LU)` goes
30,778 → 35,250 (fill 4.05x → 4.64x) on the synthetic fixture. Part A trades fill
for factorization speed — as #160 documented on its own basis
(147,796 → 172,449) — so each solve costs slightly more in absolute terms.

## Also in this PR: the two loose ends from #160's review

**`examples/basis_refactor.rs` reported population, not sample, standard
deviation.** The divisor `n.max(2.0 - 1.0)` parses as `n.max(1.0)` — `n` for any
real input — so the Bessel correction was absent. Those `±` are published, in
#160's body and in `CHANGELOG.md`. At the 15 reps that run used the factor is
`sqrt(15/14) = 1.0351`, so `9.837 ± 0.285` is really `9.837 ± 0.295` and
`0.851 ± 0.036` is `0.851 ± 0.037`.

The changelog is corrected **analytically and labelled as such**, not
re-measured: the code computed `sqrt(Σ/n)` where it should have computed
`sqrt(Σ/(n-1))`, so true = reported × `sqrt(n/(n-1))` exactly. Re-running was not
an option — the QPLIB basis behind those numbers is not in this repository. Only
the `±` move; the means, the speedups and every conclusion are untouched. The
typo is not copy-pasted anywhere else.

*Why it could ship:* `cargo test` compiles examples but does not run `#[test]`s
inside them, so a benchmark harness's own arithmetic was unguarded. CI gains a
`cargo test --examples` step (verified clean before the step was added), and
`mean_sd` gains tests against the textbook sample `[2,4,4,4,5,5,7,9]` — population
sd exactly 2.0, sample sd `sqrt(32/7)`. Those differ by precisely the error in
question, so the test fails against the pre-fix code.

**`used_dense_bump` carried `factor_nnz`'s doc comment**, and `factor_nnz` had
none. A stray line, but it ships in the public rustdoc, where the first line is
the summary shown in the method list. Moved; both now have real docs.

## Python

Issue #161's reproducer is Python, so a Rust-only fix is not reachable from it.
`LuFactor.ftran_sparse(rows, vals) -> (rows, vals)`, `btran_sparse`,
`last_sparse_solve_work`, and `hyper_sparse_max_density` on the constructor.

## Known limits

- **No real bases in-tree.** The cap defect above was invisible to the synthetic
  generator and was caught only because @jkitchin had real data. Until
  `QPLIB_1157_basis` and `QPLIB_3852_basis` are carried here as fixtures, no test
  in this repo can catch a regression of that kind. The offered attachment did
  not arrive, and they are not in discopt's git history either. **This is the
  highest-value follow-up.**
- **The eta replay is `O(eta ops)`, not `O(work)`.** Every op is walked whether
  or not its source is nonzero. Skipping zero sources is two lines and was
  deliberately not done: it diverges from `FtEta::apply_forward` on a non-finite
  multiplier (`mult * 0.0` is `NaN`) and changes no asymptotics.
- **The reach is sorted, not merged** — `O(r log r)` where a dense sweep is
  `O(r)`. This is what the 0.10 cap exists to avoid paying. Replacing the sort
  with a mark-filtered sweep in triangular order would make the 10–25% band win
  instead of lose, and would be the principled fix rather than a lower cap — but
  it is unmeasured, and shipping an unmeasured optimization to fix a measured
  regression is the wrong order.
- **No corpus bench comparison.** This container has no corpus (every partition
  `N/A`). These changes are confined to `src/lu/`, which the LDLᵀ bench does not
  exercise — an argument for low risk, not evidence of no regression.

## Tests

- `tests/lu_hyper_sparse.rs` (11) — reach vs dense route on unit rhs; dense-rhs
  fallback interleaved with sparse solves; a Forrest–Tomlin update chain;
  scaling; the zero cap; zero rhs; parameter validation; composition with the
  dense-bump route; the pinned default.
- `tests/lu_sparse_rhs.rs` (10) — agreement with the dense entry point on unit
  and multi-nonzero right-hand sides; **work-does-not-grow-with-`m`** across an
  8x range; accumulator cleanliness after failed solves; duplicate-index
  accumulation; scaling; an update chain; empty right-hand sides; independence
  from `hyper_sparse_max_density`; composition with the dense-bump route.
- 3 Python tests, 2 example unit tests.

**88 binaries, 0 failures.** Python: 68 passed, 1 skipped. `cargo fmt` and
`cargo clippy --all-targets -- -D warnings` clean on both crates, under the same
toolchain CI uses.

Reference: J. A. J. Hall and K. I. M. McKinnon, "Hyper-sparsity in the revised
simplex method and how to exploit it", *Computational Optimization and
Applications* **32**(3), 259–283 (2005).


jkitchin:
## Transfer test on the real bases — it fires and is bit-exact, but the shipped cap costs 1.45x on `btran`

This PR says the fixture is synthetic and "the transfer is **unverified**"
because the QPLIB `.npz` bases live in the discopt tree. I have run that transfer
test. The bases and the harness are attached (`feral-bases-1008.zip`:
`QPLIB_1157_basis` and `QPLIB_3852_basis` in both `.npz` and Matrix Market, plus
`real_basis_hyper.rs`, which uses the `read_mtx` helper already in
`examples/basis_refactor.rs` and needs an `[[example]]` stanza in `Cargo.toml`).

Headline: **the mechanism transfers.** The route fires on real bases and is
numerically identical to the dense sweep. But the default cap is wrong for one of
these two bases, in a way the synthetic fixture structurally cannot show.

### The route is not vacuous

`hyper_sparse_sweeps()` is asserted and the example exits non-zero if it reads
zero, because a silent fallback plus a wall-clock table is exactly how a no-op
reports success. It fired 2370–2550 of 2560 solves, `max |off − on| = 0.000e0`
on both bases, `factor_nnz()` identical across arms.

```
QPLIB_3852  m=1760  2.27 nnz/col  fill 1.01x
  ftran(e_i) solution nnz: p50=1 p90=26   max=1141 mean=27.2   (1.5% of m)
  dense ftran 1.74x    dense btran 1.02x    ftran_sparse 92x vs reach-on

QPLIB_1157  m=3937  7.46 nnz/col  fill 6.76x
  ftran(e_i) solution nnz: p50=1 p90=3593 max=3621 mean=580    (14.7% of m)
  dense ftran 10.60x   dense btran 0.67x    ftran_sparse 103x vs reach-on
```

QPLIB_3852 is the instance this PR's `lp_basis()` docstring names as the source
of its ~2.3 nnz/col target, and it behaves as the fixture predicts. QPLIB_1157 is
**4x denser per column** and does not.

### `hyper_sparse_max_density = 0.25` is a `btran` regression on QPLIB_1157

Sweeping the cap, everything else identical:

| cap | dense ftran | dense btran | route fired |
|---:|---:|---:|---:|
| 0.02 | 10.24x | 0.98x | 1195 |
| 0.05 | 10.59x | 0.98x | 1195 |
| 0.10 | 10.81x | 0.97x | 1195 |
| **0.25 (shipped)** | 11.01x | **0.69x** | 2370 |
| 1.00 | 10.75x | **0.69x** | 2560 |

The PR's own sweep found "the win is flat from 0.05 to 1.00", and that is exactly
right — **for `ftran`**. `btran` is not flat: it falls off a cliff between 0.10
and 0.25, and the fired count nearly doubles across the same step. Those extra
~1175 solves are `btran` reaches sitting between 10% and 25% of `m` that now take
the route and lose to it — the PR's own stated limitation ("the reach is sorted,
not merged — `O(r log r)` where a dense sweep is `O(r)`") biting a population the
fixture does not contain.

This is only visible on a basis whose solution-density profile straddles the cap
(QPLIB_1157: mean 14.7% of `m`). On QPLIB_3852, at the density the fixture was
tuned to, `btran` is 1.02x and the cap never bites. That asymmetry is the
concrete argument for carrying these two files in-repo rather than the synthetic
generator.

**Suggested default: `0.10`.** It holds `ftran` at 10.81x — statistically
indistinguishable from 0.25's 11.01x — and moves `btran` from 0.69x to 0.97x.

### Downstream reality check

Measured through discopt (the consumer that motivated #161), on the root LP of
QPLIB_1157 with RLT on, discopt held fixed and only feral varied:

| feral arm | in-house wall | vs crates.io |
|---|---:|---:|
| 0.15.1 (crates.io) | 6.088 s | 1.00x |
| `main` + `dense_bump_max_dim = 4096` | 3.567 s | 1.71x |
| this PR + bump + cap 0.10 | 3.607 s | 1.69x |

**Through the dense entry points this PR is ~1.00x end-to-end on this instance**,
despite the measured 10.8x on `ftran`. Triangular solves turn out to be ~1% of
that LP's wall — I have retracted the "93.1% of wall in the LU layer" claim from
#161 accordingly; that figure was a profile-attribution artifact and this PR was
built on it. On three other QPLIB instances the PR is worth 4–13% over the same
baseline, so it is not nothing, but the 476x should not be read as an end-to-end
prediction.

None of that is an argument against merging — the mechanism is sound, bit-exact,
and free. It is an argument for (a) fixing the cap default before merge and (b)
not sizing follow-up work off the isolated speedup. The `ftran_sparse` /
`btran_sparse` half remains untested from my side: exercising it needs call-site
changes in discopt, and given the ~1% attribution above I cannot currently
justify that work.

### Caveat on the numbers

Wall-clock figures come from a machine at load 4.6–6.4. Interleaved arms,
alternating order, and a flat control argue the comparisons are sound, but treat
absolute microseconds as indicative. The fired counts and the `work` counters are
deterministic and are not subject to this.

jkitchin:
Default lowered to `0.10` in 5c1d537, and the PR body rewritten so it no longer
rests on the retracted framing. Thank you for running the transfer test — both
findings are things I could not have reached from here, and the first is a
defect in my work rather than a tuning preference.

## The cap

Adopted your `0.10` and pinned it with
`default_density_cap_is_pinned_to_the_measured_value`, which carries your table
and states that raising it needs a basis whose solutions live in the admitted
band and *win* there. The parameter's doc comment carries the same table, since
that is where the next person will look.

Two things I want to be explicit about rather than quietly fix:

**My sweep was not wrong, it was blended.** I reported "the win is flat from 0.05
to 1.00" from one ratio computed over `ftran` and `btran` together. Your table
separates them and the cliff is obvious. Sweeping a threshold and reporting a
single blended number can hide a regression in half of what it covers, and that
is a methodology error, not bad luck with a fixture.

**The fixture could not have caught it.** `lp_basis()` produces solutions that
are either tiny (p50 = 3) or ~9% of `m` — bimodal, with nothing in the 10–25%
band that QPLIB_1157's 14.7% mean sits inside. I tuned the generator to
QPLIB_3852's 2.27 nnz/col, which is exactly the basis where `btran` is 1.02x and
the cap never bites. So the generator was tuned to the instance that cannot
falsify the parameter. That is the argument for carrying the real bases in-repo,
and I have put it at the top of the follow-up list.

I adopted the change **on your measurement, not reproduced** — see below. It is
the conservative direction (a smaller cap strictly reduces the population taking
the new route, so the worst case is forgoing a win rather than introducing a
loss), which is why I was comfortable taking it unverified.

## The retraction

This matters more than the cap and I have propagated it rather than footnoted it.
The PR body now leads with a pointer to the end-to-end section, labels the 476x
as a component ratio, and names numeric factorization as the LU cost that
actually moves that LP. Changelog and `dev/research/hyper-sparse-solves-2026-08-13.md`
likewise.

The part I should own: I built the `ftran_sparse` half of this on "93.1% of wall
in the LU layer" and never treated it as a hypothesis. I had no way to check it
from this container, which is exactly when a number should be quoted as *the
issue reports* rather than as the premise. A 10.6x speedup that moves the total
by 1% is a clean measurement that the component was 1%, and I agree the
`ftran_sparse` follow-up should not be sized off the isolated ratio. Declining
the discopt call-site work on that evidence looks right to me; the API costs
nothing to carry until a consumer whose profile actually shows the solves turns
up.

## The attachment did not arrive

`feral-bases-1008.zip` is named in your comment but there is no attachment link
in the body — I fetched both comments through the API rather than reading only
the notification, and checked the PR review threads and the repo's branches too.
I also checked whether they were recoverable from discopt and they are not: the
path #161 names (`scratchpad/issue1008/`) does not exist, the real
`scratchpad/i1008/` holds 18 Python harnesses and `HYPOTHESIS.md` but no `.npz`
and no `feral_repro.py`, nothing on `perf/1008-lp-simplex-throughput` either, and
`capture.py` cannot regenerate them (hardcoded macOS worktree path, a QPLIB
corpus under `~/Dropbox`, and it captures LP relaxations rather than simplex
bases).

So `real_basis_hyper.rs` is not wired up and the `[[example]]` stanza is not
added — I did not want to commit a harness for fixtures that are not here. If you
re-attach, or push the four files to a branch, I will add the example, the
`Cargo.toml` stanza, and a gated in-tree test so the cap has a real regression
guard instead of a pinned constant.

## Everything else

`88` binaries / `0` failures, Python `68 passed / 1 skipped` against a rebuilt
wheel, fmt and clippy clean under the toolchain CI uses. The one thing in this
PR that remains entirely untested by either of us is `ftran_sparse`/`btran_sparse`
against real data — correct and work-proportional by construction and by in-tree
differential tests, but never run on a real basis. Worth weighing in the merge
decision given it is new public API.

---

jkitchin:
The zip came through. Fixtures landed, your finding reproduced — **and the
mechanism I gave for it was wrong**, which changes what the cap is.

## Reproduced

Both formats verified to describe the same matrices (`max |npz − mtx| = 0.0`,
dims and nnz as your README tabulates). `real_basis_hyper.rs` in unmodified with
its `[[example]]` stanza. Re-running your sweep here:

| cap | ftran | btran | fired |
|---:|---:|---:|---:|
| 0.02 | 7.68x | 0.96x | 1195 |
| 0.05 | 7.44x | 0.97x | 1195 |
| **0.10** | 7.18x | **0.96x** | 1195 |
| 0.25 | 7.84x | **0.62x** | 2370 |
| 0.50 | 7.50x | **0.62x** | 2370 |
| 1.00 | 7.95x | **0.61x** | 2560 |

Fired counts identical to yours, and the density profile matches to the digit
(p50 = 1, p90 = 3593, max = 3621, mean = 580.4 = 14.7% of `m`). Wall-clock
differs — this box is slower on `ftran`, and the `btran` regression is *worse*
here at 0.62x — but the cliff is in the same place. `0.10` is now confirmed
rather than adopted on trust.

## The mechanism was not the sort

Your comment and my PR both attributed this to the `O(r log r)` sort, quoting my
own "known limits" list. That was wrong, and it took one measurement to find out:
sorting 400 / 1000 / 2000 positions costs **2.8 / 9.0 / 18.3 us** against a
~105 us per-solve regression. At most a sixth of it.

Instrumenting the reach to count graph edges walked (temporary, reverted),
QPLIB_1157 `btran`:

| cap | fired | reach nodes/sweep | **DFS edges/sweep** | vs `nnz(L) = 75,084` |
|---:|---:|---:|---:|---:|
| 0.10 | 21/256 | 2 | 3,662 | **5%** |
| 0.25 | 256/256 | 584 | 68,925 | **92%** |

**The reach DFS traverses the same factor entries the numeric sweep then
traverses again.** At 0.25 it walks 92% of `nnz(L)` just to decide which columns
to visit, and then visits them. ~1.9x paid to avoid 1x, which predicts ~0.53x
against a measured 0.62x.

This reframes the cap rather than just re-deriving it. I had it filed as a
workaround for an implementation detail a cleverer sweep could remove — I had
even written up "replace the sort with a mark-filtered sweep" as the principled
fix that would let the 10–25% band win. It would not have. Finding the reach is
the same order of work as the solve, so the route can only win when the reach is
a small fraction of the whole. That is intrinsic to reach-based solving, not a
defect in this implementation, and `0.10` is the design rather than a patch.

Corrected in the parameter doc, changelog, research note, and the pinning test.

## Guarded in-tree now

`tests/data/lu_bases/` (496K, the two `.mtx`; `.npz` omitted as verified-identical
and unreadable from Rust) and `tests/lu_real_bases.rs`:

- **`default_cap_excludes_the_band_where_btran_reaches_are_near_dense`** — asserts
  on reach *sizes*, which are deterministic, not on timings. Reference caps are
  fixed at 0.02/0.25 rather than derived from the default, so raising the default
  fails the assertion about the default instead of the fixture-sanity one. I
  verified it fails with the intended message when the default goes back to 0.25.
- **`sparse_entry_points_agree_with_dense_on_real_bases`** — this closes the gap I
  flagged in the merge decision. `ftran_sparse`/`btran_sparse` now run over
  *every* column of both bases against the dense oracle. That API is no longer
  validated only against a fixture tuned to it.
- **`reach_route_is_bit_identical_to_the_dense_sweep_on_real_bases`** —
  `assert_eq!` on the `f64` vectors rather than a tolerance.

844 tests across 89 binaries, 0 failures; fmt and clippy clean.

## Not addressed here

Your two points from the #161 comment are both outside this PR and I have not
touched them: whether `dense_bump_max_dim: 0` is the right default (your data
says the route ships inert and is the large win, with `4096` regressing
QPLIB_2017 to 0.80x — that wants the panel you describe, not a flip), and the
fill / refactorization-frequency questions. Happy to take either as separate
work.

---

jkitchin:
Issue #163 is now fixed on this branch (`83dd752`), as requested. Full write-up in [#163](https://github.com/jkitchin/feral/issues/163#issuecomment-5285518041); the short version, because it changes what this PR claims:

**`SparseLuSymbolic::analyze` returns to whole-basis AMD.** The Suhl–Suhl peel from #160 moves to `analyze_triangularized`, and `LuParams::dense_bump_max_dim` now documents and tests that constructor as its prerequisite — the cap is an inert no-op alongside any other. Python's `LuFactor` has no symbolic handle, so `dense_bump_max_dim > 0` selects the peel there.

Two things worth flagging:

1. **The peel is not the less accurate ordering, and I had claimed otherwise.** I first wrote the failure up as the peel forcing structural pivots that TPP would have chosen by magnitude, then measured it: every basis from the failing discopt run, re-factored both ways. Backward error ~1e-16 under both on all 30; forward error up to 2.6e-11 with the peel's ratio to AMD at 0.0x–1.0x, never worse. It is trajectory sensitivity, not instability. The rustdoc, research note and changelog were corrected to say this — it is the third time today I asserted a mechanism that explained a symptom without testing that it *was* the mechanism, and it is noted as such in the checkpoint.

2. **So the revert rests on cost-benefit, not correctness.** The peel's standalone result on QPLIB_1157 is *more* fill (197,937 vs 190,654) for 1.04x on time; its real 4.28x is the dense-bump route, which is off by default. A change buying ~nothing does not get to perturb a downstream solver's arithmetic. Both are now behind the same door.

Nothing in this PR's solve-side work (the reach route, the sparse-rhs API, the `hyper_sparse_max_density` default of 0.10) is affected — those measurements were all taken at `dense_bump_max_dim = 0`, i.e. with the route the peel enables already off.

New here: `tests/lu_default_ordering.rs` (contract test — the measurement says there is no numerical difference to assert), `tests/data/lu_bases/bchoco06_illcond_basis.mtx`, and `examples/probe_illcond_ordering.rs`, which runs on `tests/data/lu_bases` with no discopt checkout.

Gates: 90 test binaries green, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --examples` green, `cargo fmt --check` clean. Bench unchanged (dense 2/2 worst residual 1.14e-15, sparse 2/2 worst 1.26e-16 — still the two-matrix smoke set, the corpus is absent in this container).

One unfavorable note carried into the checkpoint: `python/` has 48 **pre-existing** `cargo clippy -- -D warnings` failures (`useless_conversion` and friends from pyo3 under clippy 1.97). Not introduced here and not a CI gate — `python/` is outside the root workspace and is covered by `python-wheels.yml` — but it means the binding change in this commit was verified by `cargo check` and the wheel build rather than by clippy.

---

jkitchin (commented):
Reviewed at **895ef65**. (A note on my own process: my first pass was against a
stale worktree at 6682f48 — the branch had moved by 27 files / +38k lines,
including all of issue #163 — so anything I said about this branch before this
comment was about a different tree. In particular I withdraw an earlier claim
that the Python `LuParams` change was a positional break: the pyo3 signature has
`*` immediately after `matrix`, so every added parameter is keyword-only.)

**Verdict: approve and merge.** The code is right, the tests are the right tests,
and the change is a real speedup for the callers it was designed for. Nothing
below blocks the merge, but findings 1 and 4 are each worth a follow-up issue —
and finding 4 turned out to be the largest measured effect on this branch, in
either direction.

## Verification

- `cargo test --release` at 895ef65: **90 test binaries, 848 passed, 0 failed, 21 ignored.**
- `gh pr checks 162`: `check`, `stress-smoke` and the three wheel jobs all pass.
- End to end against a real caller: discopt's dual simplex on 14 QPLIB
  relaxation LPs, three arms interleaved **inside one binary** (one feral, one
  discopt, one md5) so the arms are seconds apart rather than separate builds:
  - `off` — dense entry points, `hyper_sparse_max_density = 0` → pre-#161 behaviour
  - `reach` — dense entry points, cap `0.10` → this PR's change 1, no caller change
  - `api` — caller rewritten onto `ftran_sparse`/`btran_sparse` → this PR's change 2

  Objectives agree to 9 decimal places in every arm on every instance, so all
  three arms walk the same pivot sequence and the wall-clock comparison is
  apples to apples.

## What is right

- All four reach traversals use a valid topological order — ascending position
  for `L`, descending for `Lᵀ`, descending/ascending `uperm` **rank** for
  `U`/`Uᵀ`. Rank and not index is the subtle part once an FT update has permuted
  `U`, and it is correct here.
- `HyperWork`'s "all zero between calls" contract is restored on every exit path
  including the error paths, and `tests/lu_sparse_rhs.rs` interleaves failing
  solves with succeeding ones to prove it rather than asserting it in a comment.
- `reach_route_is_bit_identical_to_the_dense_sweep_on_real_bases` is exactly the
  right test to have written. It is what makes the `reach` arm safe to turn on
  under a downstream solver without re-qualifying its trajectory, and I confirmed
  it end to end.
- The 0.25 → 0.10 default is backed by the measurement that actually decides it
  (reach edges vs `nnz(L)`) on a real basis, with the reason the in-tree
  generator could not have found it stated plainly.

## Findings

### 1. `ftran_sparse`/`btran_sparse` have no density guard, and it costs more than the PR assumes

This is the main finding. The docs are explicit that it is a design choice —
"When the solution is *dense* this is slower than `SparseLu::ftran` … Route on
what the caller knows about its right-hand sides." The problem is that on the
motivating caller, a dual simplex, that knowledge does not exist: the density of
`α = B⁻¹A_q` is not knowable until after the solve that produces it.

Measured, 14 instances, best of 2 reps, `api` versus `off`:

| instance | mean density of `α` | `api/off` |
|---|---|---|
| QPLIB_3775 | 1.21% | **1.359** |
| QPLIB_3650 | 1.55% | 1.336 |
| QPLIB_3642 | 1.24% | 1.329 |
| QPLIB_3525 | 2.21% | 1.269 |
| QPLIB_0752 | 4.13% | 1.108 |
| QPLIB_2353 | 6.35% | 1.106 |
| QPLIB_3841 | 9.18% | 1.089 |
| QPLIB_10044 | 6.57% | 1.076 |
| QPLIB_2738 | 7.53% | 1.043 |
| QPLIB_10063 | 2.45% | 1.019 |
| QPLIB_10054 | 14.40% | 0.962 |
| QPLIB_3413 | 38.94% | 0.850 |
| QPLIB_2022 | 21.20% | 0.845 |
| QPLIB_2055 | 61.95% | **0.711** |

log-log Pearson r(density, speedup) = **−0.944**, n = 14. Split at feral's own
shipped cap: `α` under 10% dense → **1.167x** geomean (n=10); `α` at or above 10%
→ **0.837x** geomean (n=4). Overall geomean **1.061x**.

So every instance that regresses is an instance feral's own
`hyper_sparse_max_density` default would already have routed to the dense sweep —
if the cap applied to the sparse entry points. Taking `max(1, speedup)` per
instance, an ideal guard would give **1.116x** geomean, i.e. the missing guard is
costing about half of what the API is worth.

The fix looks nearly free: the reach is computed inside the solve, so by the time
the cost becomes knowable the information is already in hand. When
`order.len() > hyper_sparse_max_density * m`, fall back to the dense sweep and
emit the nonzeros. As shipped, every caller has to rediscover this the way I did.

### 2. The sparse-rhs API is *not* bit-identical to the dense entry points, and the docs do not say so

`sparse_entry_points_agree_with_dense_on_real_bases` asserts `< 1e-9`, while
`reach_route_is_bit_identical_to_the_dense_sweep_on_real_bases` gets `assert_eq!`.
That difference is deliberate and fine, but it is invisible from the doc comment,
which is otherwise very complete. This same branch carries issue #163, whose
whole lesson is that a rounding-trajectory change with no standalone payoff cost
a downstream solver a dual bound on an ill-conditioned LP. A caller migrating
`ftran` → `ftran_sparse` is taking exactly that risk and should be told so in one
line — the more so because, per finding 1, the migration is the part with the
regression risk while the reach route is the part that is provably safe.

### 3. `SingularBasis { column }` names an internal index in the solve path

`sparse_hyper.rs:389,391,420,422` report the permuted pivot position `k`, whereas
`sparse_factor.rs` reports `qcol[k]`, the original basis column. This is
pre-existing in `sparse_solve.rs`, but the new *public* API inherits it, and
a91519d fixed exactly this class of defect on the factor path. Either map through
the permutation or document that the solve path's `column` is a permuted index.

### 4. #163's revert costs 4.3x-12.4x on the analysis, and nothing in the PR says so

This is the largest measured effect on this branch, in either direction, and I
only found it because a kernel share looked wrong.

`SparseLuSymbolic::analyze` is re-run on **every** refactorization by a simplex,
so its per-call cost is multiplied by the refactorization count rather than
amortized. Measured in feral alone, 20 reps, release, no caller involved
(`analyze` vs `analyze_triangularized` on the branch's own fixtures):

| basis | m | nnz | `analyze` | `analyze_triangularized` | ratio |
|---|---|---|---|---|---|
| QPLIB_1157 | 3937 | 29376 | 10.329 ms | 0.834 ms | **12.4x** |
| QPLIB_3852 | 1760 | 4003 | 0.247 ms | 0.038 ms | 6.6x |
| bchoco06_illcond | 833 | 2404 | 0.148 ms | 0.034 ms | 4.3x |

End to end this dominates everything else on the branch. Same 14 instances, same
binary, four arms — `api` is the sparse-rhs API, `tri` is just calling
`analyze_triangularized` instead of `analyze`, `all` is both:

| arm | geomean | median | max |
|---|---|---|---|
| `api` | 1.067x | 1.090x | 1.363x |
| `tri` | **1.306x** | 1.469x | 1.674x |
| `all` | **1.497x** | 1.729x | 2.584x |

On QPLIB_3775, one instance in full: `LuSymbolic` 1048.5 ms across 64
factorizations versus `LuNumeric` 184.6 ms. The branch spends **5.7x the numeric
factorization on choosing a column order**, 64 times over. Switching the ordering
takes that LP from 2.215 s to 1.193 s with an identical objective and an
identical factorization count.

I want to be clear about what this is and is not. It is **not** an argument to
revert #163 — #163 is right that the peel is a trajectory change and that a
caller has to opt into it. QPLIB_2055 makes the same point from the other side:
under `tri` it is **0.389x**, a 2.6x slowdown, and its objective moves in the 9th
significant figure, so the peel changed its pivot path too. The peel is not free
and it is not universally good.

What is missing is that the revert's *cost* was never measured. The `analyze`
doc comment argues the accuracy case at length and says nothing about the 4.3x-12.4x
it hands back, and the peel is now reachable only by a caller who reads the module
docs closely enough to find `analyze_triangularized`. Two suggestions, neither a
blocker:

- Put the cost in the `analyze` doc comment next to the accuracy argument, so the
  tradeoff a caller is being asked to make is stated in both directions.
- Consider making the choice a `LuParams`/analysis knob with a documented default
  rather than two separately-named constructors, so it is A/B-able by callers
  without a code change — which is exactly how I had to measure it.


### 5. Scope note

The title says "reach-limited sweeps + sparse-rhs API", but the branch also
carries issue #163, which reverts #160's headline default (`analyze` no longer
peels). The revert is defensible and unusually well documented — but it is a
behaviour change to the previous PR's main feature and it is invisible from the
title.

## One thing to chase separately

QPLIB_3225 solves under a feral built from `main` but **fails to solve under this
branch in all three arms, including `off`** — so it is not the reach route and not
the sparse API. The only remaining feral difference is #163's ordering revert,
which would mean the revert bought bchoco06 and lost QPLIB_3225. I have not
isolated it under a controlled arm=main/arm=branch rerun yet; flagging it rather
than asserting it. It should not hold up this merge, but it is worth an issue.

jkitchin:
Thank you for this, and particularly for finding 4 — it is a correction to my work, not a suggestion, and it was measurable from inside this repository the whole time. Addressed in `12d1fcc`; findings 1, 4-ergonomics and the QPLIB_3225 lead are filed as #164, #165, #166.

## Finding 4 — my rationale was wrong, and I verified that before rewriting it

Reproduced with `examples/basis_refactor.rs`, 20 reps, release, on all three in-tree fixtures:

| basis | `analyze` | `analyze_triangularized` | symbolic | total |
|---|---|---|---|---|
| QPLIB_1157 | 21.280 ± 0.895 ms | 2.166 ± 0.176 ms | **9.83x** | 1.03x |
| QPLIB_3852 | 0.857 ± 0.083 ms | 0.130 ± 0.018 ms | **6.61x** | **2.73x** |
| bchoco06 | 0.536 ± 0.044 ms | 0.127 ± 0.014 ms | **4.22x** | **2.26x** |

Your ratios reproduce. Two compounding errors on my side: I timed only numeric factorization — `lu_fill_orderings.rs` does not time symbolic, while `basis_refactor.rs`, which I had already used that session, times both — and then generalized 1.03x from the single fixture where the peel's *total* advantage is smallest, with 2.73x and 2.26x sitting in the same directory.

The worst part is that the disconfirming number was already in `CHANGELOG.md`'s #160 entry, which I edited in the same session: "cutting the ordering from 9.837 ± 0.295 ms to 0.851 ± 0.037 ms." That is your 9.8x, in a file I was writing to.

**The decision survives on a weaker argument and every document now says so.** Not "the peel is free to give up" but "neither ordering dominates, so the caller must choose": for it, the speedups above plus being the precondition for `dense_bump_max_dim`; against it, #163's lost dual bound and QPLIB_2055 at 0.389x with the objective moving in the 9th significant figure. `analyze` stays whole-basis AMD because that is the trajectory your suite was green against — not because it is faster or more accurate, which it is not, reliably. Corrected in the `analyze` / `analyze_triangularized` / `dense_bump_max_dim` rustdoc, the module header, the Python doc, `CHANGELOG.md`, and the research note. `decisions.md` is append-only, so the correction is a new entry naming the one it corrects.

Your second suggestion — a parameter with a default rather than two constructor names — is #165. I did not do it here: it does not fit `LuParams` (the ordering is chosen at symbolic time, `LuParams` is consumed at factor time), so it needs an API-shape call that is yours to make. I folded AMD-vs-AMF into that issue, since it is the same "instance-dependent lever that isn't A/B-able" problem — AMF is 21% less fill and 2.10x numeric on QPLIB_1157 and isn't wired into the LU at all.

## Finding 3 — fixed, and the old tests could not have caught it

All four sites now map through `qcol`. The existing L10 tests are 2×2 with an identity ordering, so position and column coincide there. The new test uses a 6×6 whose AMD gives `qcol = [5,4,0,1,2,3]`, asserts its own non-vacuity (`assert_ne!(want, pos)`), and covers all four entry points. Confirmed it fails without the fix:

```
ftran must name original column 4, got Err(SingularBasis { column: 1 })
```

I fixed the dense kernels too, not just the new API — the pre-existing behavior was wrong in the same way and a91519d set the precedent.

## Finding 2 — added, with the concrete hazard named

The `ftran_sparse` doc now says it is not bit-identical to `ftran`, and says why that matters in the specific form you identified: migrating is a rounding-trajectory change, which is exactly what cost a solver its dual bound in #163. It also points callers who want the speedup without the trajectory change back at the dense entry points, whose reach route *is* pinned bit-identical.

## Finding 1 — filed as #164, not implemented

Your data is decisive and I'd have taken the change on it. I stopped because it changes documented semantics of the new API (`prepare_sparse_solve` promises the sparse entry points work *regardless* of `hyper_sparse_max_density`) and the test that pins that promise, you scoped it as follow-up-worthy, and there are three second-order questions worth your call rather than mine — whether the fallback shares the existing knob or gets its own, and what it should do to `last_sparse_solve_work()`, which currently exists to prove there is no `O(m)` term.

What I did instead: put your whole table in the `ftran_sparse` rustdoc, including the part I think is the actual finding — that a dual simplex **cannot** route on density, because `B⁻¹A_q`'s density isn't known until the solve that produces it has run. "Route on what the caller knows" was written without checking whether the motivating caller could know.

## Finding 5 — title updated

Now "work-proportional sparse LU solves (#161B), plus the #163 ordering revert".

## Gates

90 test binaries, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --examples` green, `cargo fmt --check` clean.

One thing I have not resolved and am flagging rather than burying: `python/` carries 48 **pre-existing** clippy `-D warnings` failures (`useless_conversion` and friends from pyo3 under clippy 1.97). They are outside the root workspace so the hook never sees them, which is presumably how they accumulated. The Python doc change in `12d1fcc` was verified by `cargo check` and the wheel build, not by clippy.

---

jkitchin:
Pushed the two follow-ups this review asked for, plus the measurement that
justifies the first one. Head is now 255e2b7.

- **7f1682e — issue #164, the density guard.** `LuParams::sparse_rhs_max_density`
  (default `0.10`): past that fraction of `m`, the reach DFS is abandoned
  mid-walk and the kernel sweeps the whole basis instead. Details and the shape
  argument are in jkitchin/feral#164.
- **4a83bab — issue #165, the ordering parameter.**
  `SparseLuSymbolic::analyze_with(a, LuOrderingParams { triangularize })`, both
  existing constructors kept as wrappers, default unchanged. Details in
  jkitchin/feral#165. The AMD-vs-AMF half of that issue is **not** done — AMF
  is not wired into the LU at all.
- **b8906a6** — the session-03 checkpoint carrying this review.
- **255e2b7** — the measurement below.

## What the guard is worth, measured

Same design as the review's panel: four arms interleaved in ONE binary (one
feral, one discopt), 14 QPLIB relaxation LPs, 2 reps, best-of. The `api` arm is
split so the new cap is the only difference between two of them.

|  | geomean | median | min | max |
|---|---|---|---|---|
| `apiX/off` — the branch as reviewed | 1.066x | 1.083x | 0.735x | 1.405x |
| `api/off` — with the #164 guard | **1.197x** | 1.196x | **0.998x** | 1.408x |
| `api/apiX` — the guard alone | 1.122x | 1.122x | 0.929x | 1.393x |

The reviewed arm reproduces (1.066x now against 1.061x then). Every regression is
gone: the worst instance on the panel is now 0.998x, i.e. parity, and the four
dense-`alpha` instances go 0.839x -> 1.067x as a bucket. QPLIB_2055, the extreme
at 61.95% dense, goes 0.735x -> 1.024x.

This is **better than the 1.116x the review predicted**, and the review's own
reasoning is why the prediction was low: `max(1, speedup)` per instance assumes a
guard can only claw the regressing instances back to parity, but the density
column is a *mean* over the solve while the guard routes *per solve*. Instances
whose mean `alpha` is nominally sparse still have dense individual solves, and
those get caught — the under-10% bucket improved too, 1.174x -> 1.253x. The
mean-`alpha` column also says nothing about the btran side, which the guard
covers as well.

Objectives are bit-identical across `off`, `apiX` and `api` on all 14 instances.
Stating that as the observation it is, not as a guarantee: the tests still assert
`< 1e-9` between the routes rather than equality (this review's finding 2), since
the two accumulate in different orders and are entitled to differ in the last
bits. The only arm that moves an objective is the peel arm, on 8 of 14 instances
— the known #163 trajectory effect, and not on by default.

`cargo test --release`: 90 binaries, **855 passed, 0 failed**, 21 ignored (848 at
895ef65; the +7 are the new tests). `cargo clippy --all-targets -- -D warnings`
and `cargo fmt --check` clean in both the root crate and `python/`.

Still worth a separate issue and not addressed here: QPLIB_3225 failing under
this branch in all arms (the review's closing note).

jkitchin:
Reviewed both commits and re-ran the gates here: **90 binaries, 853 passed, 0 failed, 21 ignored** (release), `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean. One small follow-up pushed as `e00aa70`.

## The #164 guard answers all three things I held it back for

- **Its own parameter.** `sparse_rhs_max_density` rather than overloading `hyper_sparse_max_density`, so "the sparse entry points work regardless of the dense route's cap" stays true and the two caps can move independently.
- **Abandoned mid-walk**, not after `finish_reach`. I had sketched checking the reach after building it, which pays the whole DFS to discover it was not worth building. Yours pays a bounded fraction.
- **`go_dense` marks the whole accumulator** before the sweep, which is the part I could not see how to do cheaply and is why I did not attempt it. Making `reset()` `O(m)` exactly in the regime where the answer is already `Θ(m)` is the right trade and the doc says so.

The four dense branches use the correct topological orders — `0..m` for `L`, reversed for `Lᵀ`, and `uperm_inv[rank]` for both `U` sweeps rather than position order, which is the part that would have been silently wrong after an FT update.

`analyze_with` + `LuOrderingParams` is the right shape, and `analyze_with_reproduces_both_constructors_exactly` pins the thing that actually matters — that it is not a *third* ordering — including the `triangularized` provenance flag that `dense_bump_max_dim` gates on.

## One finding: the contract test did not guard its contract

I checked `the_density_fallback_restores_the_zeroed_accumulator` by deleting the line it exists to protect — `go_dense`'s `mark.fill(true)` / `pattern.extend(0..m)` — rather than by reading it.

**It passed.** The stale entries missed the one column it probed per round. The corruption was caught, but by `the_density_fallback_agrees_with_the_reach_route_and_the_dense_api` (diverged by 5.5e-2), not by the test named for the invariant. A contract test whose verdict depends on which column it happens to pick is not guarding the contract — and this one sits in front of the subtlest part of the change.

`e00aa70` sweeps a rotating quarter of the columns per round instead of one. Verified both directions: fails with the marking removed, passes with it. The implementation was always correct; only the test's grip on it was loose.

## Two notes, neither needing action

**Your `apiX/off` → `api/off` explanation is right and worth keeping.** My 1.116x estimate was `max(1, speedup)` over a *mean*-density column, which by construction cannot exceed parity on the regressing instances. The guard routes per solve, so it also catches dense solves inside nominally sparse instances — which is why the under-10% bucket moved at all (1.174x → 1.253x). The prediction was low because the model was coarser than the mechanism, not because the guard over-performed.

**Test count.** You report 855, I get 853 on the same commit — same 21 ignored, same 90 binaries, both green. Likely an environment difference rather than anything skipped, but flagging it since it is the kind of gap that occasionally means two tests are not compiling somewhere.

Also: your comment gives the head as `255e2b7`, but at the time I fetched, `origin` was at `b8906a6` — `255e2b7` and `55c0ff4` arrived while I was working, and I have rebased onto them. No action needed, just noting it in case the push was later than intended.

The QPLIB_3225 lead is filed as #166 with the isolation recipe, so it survives the merge.

---

jkitchin:
Heads-up before this merges: I could not reproduce the `peel + cap 4096` row of
the table `895ef65` added to `dev/decisions.md` — filed as #168 with the full
reproduction.

Short version: same discopt commit (`bce881ff`), only feral varying, patched the
way `895ef65` describes. Rows 1 and 2 reproduce exactly; row 3 fails.

| ordering | cap | result | dense-bump firings |
|---|---|---|---|
| whole-basis AMD | 0 | ok | 0 |
| peel | 0 | FAILED | 0 |
| **peel** | **4096** | **FAILED** | **26** |

Identical at `895ef65` itself, so it is not a regression from `4a83bab` /
`7f1682e` / `12d1fcc`. The firing counter is patched straight into
`sparse_factor.rs` after `want_dense_bump`, so the failing arm is one where the
dense route demonstrably ran — this is not the vacuity mode you flagged for
`used_dense_bump()`.

If that holds up, the conclusion "#160 can ship and the speedup can be taken"
doesn't, and #163's coupling stands: `symbolic.triangularized` is a hard
precondition at `sparse_factor.rs:334`, so there's no cap-without-peel fallback.

Not a blocker for the hyper-sparse work in this PR, which is independent and
which I have nothing against — it's the `dev/decisions.md` claim and the release
framing that need re-checking. Happy to hand over the scratch trees.

PR #191, opened 2026-08-21
fix(solve): default refinement certifies componentwise accuracy (closes #190)

### LINKED ISSUE #190: RefineOptions needs a residual target, not just a step cap
## Summary

`RefineOptions` lets a caller cap the number of correction steps but not say
what accuracy it actually wants. For an interior-point host that runs its own
refinement loop over the same system, the step cap turns out to be an unusable
lever: the target is unreachable on the matrices that matter, so the loop
always runs to the cap, and every cap value is a different arbitrary answer.

Two asks, in order of value:

1. Let the caller name the relative-residual target, not just the step budget.
2. Surface the step count / `RefinementDiagnostics` through the high-level
   `Solver::solve_refined_into` / `solve_many_refined_into` entry points, so a
   host can see whether the budget bound without dropping to
   `solve_sparse_refined_with_diagnostics_opts` and the raw factors.

This is the same shape as #178, which made the budget a per-call parameter and
which POUNCE asked for. #178 was the right change; it just turns out the budget
is the wrong knob.

## Why the step cap is not a usable lever

POUNCE (pounce gh#710) sets `RefineOptions::with_max_steps(k)` per back-solve.
Sweeping its 118-leg fixture corpus at each `k` gives this:

| `max_steps` | badly-scaled LP (data scale 1e11) | `deb7` exact-Hessian | `cresc4` L-BFGS | 58k-var benchmark, wall |
|---|---|---|---|---|
| 0  | **RestorationFailed** | ok, 146 it | ok, 105 it | 16.7 s |
| 1  | ok | **ErrorInStepComputation, 183** | **InfeasibleProblemDetected, 32** | 26.1 s |
| 2  | ok | **ErrorInStepComputation, 258** | ok, but 997 it | 33.9 s |
| 10 | ok | ok, 171 it | ok, 143 it | 61.7 s |

Each cap loses something different, and it is not monotone in `k`. One more
fixture makes the point sharper — `eigena2` under limited-memory is `Optimal`
at `k=5`, `SolvedToAcceptableLevel` at `k=10`, and `ErrorInStepComputation` at
`k` = 0, 1, 2, 3 and 4.

`cresc4` at `k=1` is the one that should worry you: a **false infeasibility
verdict on a feasible problem**, reached in 32 iterations. That is not a slower
route to the right answer, it is a wrong answer.

The current default of 10 survives the corpus, but nothing about the sweep
suggests it survives it *for a reason*. It reads as a lucky ticket.

## Why it behaves that way

On a 118,276-dimension near-singular KKT the convergence test at
`numeric/solve.rs:2574` (and `:2749` for the multi-RHS path) —

```rust
let threshold = f64::EPSILON * n_sqrt;
...
r_norm < threshold * b_norm
```

— is simply not reachable. The loop therefore runs at or near `max_steps` on
every back-solve. The 2-strike stagnation rule does not rescue it, because each
step *does* improve the best residual slightly; it just never gets to
`ε·√n·‖b‖`. The divergence guard does not fire either.

So the best-iterate contract hands back whichever of the `k+1` non-converged
iterates happened to have the smallest `‖r‖₂`. Those iterates are bouncing
around a floor they cannot cross, and smallest `‖r‖₂` **on the condensed
symmetric system** has no reliable relationship to which one gives the better
Newton step in the host's outer loop. That is the mechanism behind the table
above: each cap is close to a draw.

## What the host actually needs

The host is not trying to solve `Ax=b` accurately for its own sake. It wants
the backend to return what a stable `LDLᵀ` would have returned — a backward-
stable solve of the system it factorized — and then it does its own refinement
on top, against a *different* system.

This matters more than it sounds. Ipopt (and POUNCE, which ports it) applies
iterative refinement to the **unreduced non-symmetric Newton system**, not to
the condensed symmetric system that was factorized. Wächter & Biegler §3.10:
the condensation `Σ = S⁻¹Z` destroys information as `μ → 0`, so a residual
measured on the condensed system is blind to exactly the loss that matters. A
sparse backend can only ever refine the condensed system it factorized — the
wrong system — which is why Ipopt disables backend-internal refinement on every
direct solver it ships:

- MA27 — no such routine
- MA57 — `MA57D` is never declared or called; `IpMa57TSolverInterface.cpp:785`
  calls only `ma57c`
- MUMPS — `IpMumpsSolverInterface.cpp:428`: `mumps_data->icntl[9] = 0;  //no
  iterative refinement iterations`
- Pardiso Project — registered default 0

Only MKL Pardiso (default 1, "makes it more robust and just a little bit
slower") and WSMP (`IPARM(3)=5`) opt in.

So the *architecturally* correct setting for POUNCE is `max_steps = 0`. It
can't use it, and the reason is feral-specific: `ZeroPivotAction::ForceAccept`
is the default, so feral's raw solve can leave a non-trivial residual against
the very system it factorized — which is precisely the assumption Ipopt's
design makes about a backend and which MA27/MA57/MUMPS satisfy by doing
threshold pivoting with delayed pivots instead. That is what loses the
badly-scaled LP at `k=0` in the table. `solve.rs:2027` already says this out
loud:

> because `ZeroPivotAction::ForceAccept` is the default, an unrefined solve
> can leave a non-trivial residual on near-singular pivots

POUNCE therefore needs *some* refinement to get back to "honest solve", and
then wants it to stop. It has no way to say that. It can only say "ten steps"
or "one step" and take whatever falls out.

## Cost

58,014-variable trajectory-optimization benchmark, limited-memory Hessian,
60 iterations, same objective to the digit at every setting:

| setting | Factorization | **BackSolve** | Overall |
|---|---|---|---|
| `max_steps = 10` (default) | 9.966 s | **48.962 s** | 61.741 s |
| `max_steps = 2` | 7.867 s | 23.326 s | 33.947 s |
| `max_steps = 1` | 7.280 s | 16.093 s | 26.130 s |
| `max_steps = 0` | 6.478 s | 7.443 s | 16.728 s |

45 seconds — 73% of wall time — spent driving a residual the host never reads,
on a system that is not the one it refines against. For reference the same
problem through MA57, which does no inner refinement at all, is 10.1 s overall.

Ruled out as the cause: this is **not** the L-factor perturbation from
cascade-break. The same run with CB disarmed is 62.113 s against 61.741 s
armed, to a bit-identical objective. (Noting it because the belief that CB is
what makes the full budget reachable had propagated into POUNCE's own docs, on
the strength of #179's Hilbert / bordered-KKT measurements. Those hold; they
just don't describe this matrix.)

## Proposed API

```rust
pub struct RefineOptions {
    pub max_steps: usize,
    /// Relative-residual target: stop once `‖r‖₂/‖b‖₂` is below this.
    /// `None` keeps the current `ε·√n`.
    pub target: Option<f64>,
}

impl RefineOptions {
    pub fn with_target(target: f64) -> Self { ... }
    pub fn and_max_steps(self, max_steps: usize) -> Self { ... }
}
```

Adding a field breaks `RefineOptions { max_steps: k }` struct-literal
construction, so this wants a minor bump (or `#[non_exhaustive]` plus builders,
if you'd rather take that break once).

I'd rather not prescribe the value POUNCE should pass — with the knob in place
that becomes something we can sweep the corpus against, which is exactly what
we cannot do today. The point is that "good enough for the caller" is the
caller's number, and right now it is hard-wired to machine precision.

The second ask is smaller: `RefinementDiagnostics` and `RefinementStep` are
public, but the only way to get them is
`solve_sparse_refined_with_diagnostics_opts` with the raw factors.
`Solver::solve_refined_into` and `solve_many_refined_into` are what a host
actually calls, and they discard the step count. Returning it — even just
`usize` — would have made the "does it run to the cap?" question above a
one-line measurement instead of an inference from back-solve cost.

## Reproducing

Both halves are in `jkitchin/pounce`:

```
# the corpus sweep in the first table (both Hessian legs)
scripts/sweep-fixtures.sh <pounce-binary> out.txt feral_refine_steps=<k>

# the timing table
pounce benchmarks/large_scale/nl/laptime.nl \
    hessian_approximation=limited-memory max_iter=60 \
    print_timing_statistics=yes feral_refine_steps=<k>
```

`laptime.nl` is generated by `benchmarks/large_scale/generate_nl.py laptime`.
The `feral_refine_steps` option is POUNCE's passthrough to
`RefineOptions::with_max_steps`, added when #178 shipped.

Context on the POUNCE side: pounce gh#698 (observation 5), gh#710.

jkitchin:
## Read this first: breaking API change, and a measured cost

**Breaking.** Three `Solver` methods now return `Result<RefineOutcome, FeralError>`
where 0.17.0 returned `Result<(), FeralError>`:

- `Solver::solve_refined_into`
- `Solver::solve_many_refined_into`
- the parallel refined entry point reached from `solve_refined_into`

Any caller using the result in *expression* position — a `match` arm, a tail
expression in a function returning `Result<(), _>`, a `let () =` binding — fails
with `E0308`. pounce does not compile against this branch until its call sites
add `.map(|_| ())`. Adding `stop` to `RefineOptions` is source-compatible, so the
return type is the only break. Under 0.x semver **this makes the next release
0.18.0, not 0.17.1.**

The CHANGELOG previously claimed "existing callers compile unchanged — the new
return value is not `#[must_use]`." That was wrong: `#[must_use]` governs whether
*discarding* a value warns, not the value's type. Corrected in this branch.

**The componentwise default costs latency.** A controlled A/B over the corpus —
same binary, only the `RefineOptions::default()` stop criterion differing:

| ratio | control (`EpsSqrtN`) | shipped default | delta |
|---|---:|---:|---:|
| factor/MUMPS p90 | 1.54 | 1.56 | +1.3% (noise) |
| solve/MUMPS geomean | 0.08 | 0.08 | none |
| solve/MUMPS p50 | 0.08 | 0.08 | none |
| solve/MUMPS p90 | 0.15 | 0.20 | **+33%** |
| solve/MUMPS p99 | 0.71 | 1.08 | **+52%** |
| solve/SSIDS geomean | 0.94 | 1.06 | **+13%** |
| solve/SSIDS p90 | 2.50 | 3.60 | **+44%** |
| solve/SSIDS p99 | 8.33 | 13.00 | **+56%** |

**And on at least one real workload it is not a tail effect — it is the whole
run.** Measured downstream in pounce on `laptime` (58,014 variables, L-BFGS,
126,028-dimension KKT, 60 IPM iterations), one binary, arms differing only in the
`StopCriterion`:

| stop criterion | overall | factor | back-solve | vs `EpsSqrtN` |
|---|---:|---:|---:|---:|
| `EpsSqrtN` (0.17.0's default) | 58.6 s | 9.27 s | 46.7 s | — |
| refinement off | 13.9 s | 5.85 s | 5.44 s | -76% |
| `RelativeResidual(1e-10)` | 22.8 s | 6.48 s | 13.77 s | -61% |
| **this PR's default** | 69.4 s | 10.39 s | 56.4 s | **+18%** |

+18% overall and +21% on back-solve, on every iteration. `dirichlet120` (exact
Hessian) is factorization-dominated and does not discriminate — all four arms
reach `Optimal` in 56 iterations at the same objective.

Whether the conjunctive default is right is a reviewer decision. The alternative
is to keep `EpsSqrtN` as the default and let callers opt in — `f547bc5` makes
that work either way, and it is a one-line change to the `Default` impl at
`src/numeric/solve.rs:2140`.

---

## What the deficiency was

pounce reported that feral works on the vast majority of problems but fails on a
class of them where MA57/MUMPS succeed. Four hypotheses were ruled out by
measurement — `ForceAccept`, the pivot threshold, the escalation ladder, and
inertia. None of them was the gap.

The gap is the **stopping criterion**. Iterative refinement stopped on the
normwise residual `‖r‖₂/‖b‖₂`, which is dominated by the rows with the largest
`|bᵢ|`. When an interior-point method drives components of the RHS across many
orders of magnitude — the shape a KKT system takes near convergence — a solution
can satisfy the normwise test while individual components are badly wrong.
MUMPS stops on the componentwise backward error instead
(`CNTL(2) = sqrt(eps)`, `ref/mumps/src/dini_defaults.F:1094`), reached through
`ICNTL(10)`.

feral had no way to ask for that. Now it does, and it is the default:

```rust
StopCriterion::EpsSqrtNAndBackwardError(DEFAULT_BACKWARD_ERROR_TARGET)
```

The conjunction, not a replacement — `BackwardError(√ε)` alone regresses
`tests/parity.rs` (`ROSZMAN1_0241 residual: feral=4.805e-14 > max(K*mumps=2.077e-14, floor=1.000e-14)`),
and `BackwardError(f64::EPSILON)` stagnates or exhausts the step budget on 5 of
7 large matrices. Both are recorded in `dev/tried-and-rejected.md`.

## Evidence

`tests/issue190_componentwise_default.rs`, three tests against tracked fixtures:

- **`default_certifies_backward_stability_on_badly_scaled_rhs`** — all 63
  factorable parity matrices return `ω ≤ √ε` under a RHS built as `b = A·v` with
  `v[i] = ±10^((i%13)-6)`.
- **`normwise_criterion_alone_is_not_enough`** — pins 5 named fixtures where the
  historical rule leaves `ω > 10·√ε`: `DEGENLPB_0046` (5.35e-6), `DEGENLPB_0045`
  (4.48e-6), `DEGENLPB_0047` (2.69e-6), `DEGENLPA_0065` (1.89e-6), `ACOPP14_0003`
  (6.42e-7). Assertions are one-sided so drift cannot flip them.
- **`default_is_never_weaker_than_the_historical_rule`** — the conjunction never
  yields a worse `ω` than `EpsSqrtN` on any corpus matrix.

`ω` is computed with LAPACK `dgerfs`'s guarded formula (Arioli–Demmel–Duff 1989),
so the oracle is external to this implementation.

**Full suite:** 956 tests pass. **Inertia gate:** 154531/154590 = 100.0%.
**Phase 2.8.1 partitions:** all four PASS, at the bottom of the observed band
(sparse 1.56/1.56, dense 1.58/2.00 against the 08-19-05 baseline of
1.58/1.58/1.58/2.00).

## Also in this branch

Two independent threads that were already sitting unpushed on local `main`:

- **`a0b4d64` — multi-RHS cache-line stride.** The row-major multi-RHS work
  buffers used a leading dimension of exactly `nrhs`, so any `nrhs` not a
  multiple of 8 straddled a cache line on every panel row. `t(31)/t(32)` — where
  `nrhs = 31` is 3% *less* work — read 1.41–1.80 on seven matrices. Padding the
  panel path's row stride to `nrhs.div_ceil(8) * 8` brings it to 0.73–1.38.
  Measured gain in the regime that actually ships (`nrhs >= 32`, not a multiple
  of 8): **geomean 1.36x**. Bit-for-bit identical output, verified three ways
  including an 800-shape `to_bits()` sweep against a scalar reference.
  **This does not reach pounce**, which caps `nrhs` at 16 deliberately to stay
  under `BLAS3_NRHS_THRESHOLD` and keep the blocked solve bit-identical to
  looping (pounce gh#729). The 1.36x is real but downstream-invisible for that
  caller.
- **`f547bc5` — `StopCriterion`.** `EpsSqrtN` / `RelativeResidual(f64)` /
  `BackwardError(f64)` / `EpsSqrtNAndBackwardError(f64)`. Lets a caller state
  what convergence means instead of inheriting one hardcoded rule. This is also
  what retires pounce's `FeralConfig::refine_target` placeholder and the wasted
  unrefined pre-solve it guarded.

Plus four diagnostic probes (`001a7db`) that located the gap, and a docs commit
(`f5bc004`) removing unverifiable provenance numbers from `849caea`'s growth-flag
comments.

## Correction to an earlier claim

`crates/feral-diagnostics/src/bin/probe_vs_mumps_residual.rs` used to carry a
comment claiming feral fell "nine orders short" of MUMPS on residual. That was
wrong — it compared feral *unrefined* against MUMPS *with* refinement enabled.
Refined, feral matches or beats MUMPS on all seven large matrices under that RHS,
and inertia agrees exactly on all seven. The comment now says so.

## Open items, not addressed here

- `qap15_kkt` returns `ω = 1.229e-10` where MUMPS reaches 1.023e-12. Below the
  target, so it passes, but the gap is unexplained.
- The `residual_improvement_factor` plateau heuristic deserves its own issue.
- The `BLAS3_NRHS_THRESHOLD` doc comment still cites `multi-rhs.md` D3 as if it
  were a measurement; it is a pre-implementation rule of thumb.
- The six-file version bump to 0.18.0 (`scripts/release-checklist.sh bump
  0.18.0`) belongs to the release ritual, not this PR; the CHANGELOG now names
  0.18.0 as the target.

Closes #190.

https://claude.ai/code/session_01GdAxgLjVsYe4oB2XRURmKG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,10 +29,25 @@ All notable changes to FERAL will be documented in this file.
     of `eps` says "as good as a backward-stable factorization would have
     returned," independent of `n` and of row scaling.
 - **What you get back.** The `*_refined_into` entry points now return a
-  `RefineOutcome { steps, relative_residual, stop }` instead of `()`, where
-  `stop` is `Converged | MaxSteps | Stagnated | Diverged`. All three values
-  are already computed by the loop, so reporting them costs nothing. The
-  full `RefinementDiagnostics` is deliberately *not* returned here: it
+  `RefineOutcome { steps, relative_residual, backward_error, stop }` instead
+  of `()`, where `stop` is `Converged | MaxSteps | Stagnated | Diverged`.
+  All four values are already computed by the loop, so reporting them costs
+  nothing.
+
+  `backward_error` is the achieved componentwise `omega` of the returned
+  iterate — the quantity the new default criterion certifies, so a host that
+  gets `Converged` back can log or threshold on the certificate without
+  paying a second `abs_symv` to recompute it. For a multi-RHS call it is the
+  worst column, matching `steps` and `relative_residual`.
+
+  **It is `NaN` when `omega` was not measured** — under `EpsSqrtN`,
+  `RelativeResidual`, and `max_steps = 0`, where the loop never forms it.
+  `NaN` here means *not measured* and is deliberately distinct from
+  `INFINITY`, which is a real measurement (a non-finite iterate). Test it
+  with `is_nan()` before comparing against a bound; `NaN < t` is `false`, so
+  a bare comparison reads "not measured" as failure.
+
+  The full `RefinementDiagnostics` is deliberately *not* returned here: it
   computes a Hager-Higham condition estimate costing 3-5 extra solves, which
   would multiply exactly the cost this change exists to cut. Ask for it
   explicitly via `solve_sparse_refined_with_diagnostics` when you want it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,53 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Added - iterative refinement can be told what "converged" means (issue #190)
+
+- **The problem.** The refinement loop's only convergence test was a hardwired
+  `||r||_2 < eps * sqrt(n) * ||b||_2`. That target is not reachable on large
+  ill-conditioned systems — at `n = 118,276`, `eps * sqrt(n) = 7.6e-14` — so
+  every call ran the full `max_steps` budget and paid for iterations that
+  bought the caller nothing. The only lever was `max_steps`, a step count with
+  no principled value: the right count is problem-dependent, so tuning it is
+  guessing.
+- **What changed.** `RefineOptions` gains a `stop: StopCriterion` field:
+  - `StopCriterion::EpsSqrtN` — the historical behavior, and still the
+    default. Bit-for-bit unchanged.
+  - `StopCriterion::RelativeResidual(t)` — stop once
+    `||r||_2 <= t * ||b||_2`.
+  - `StopCriterion::BackwardError(t)` — stop once the componentwise backward
+    error `max_i |r_i| / (|A| |x| + |b|)_i <= t` (Arioli-Demmel-Duff 1989;
+    the quantity MA57 reports in `RINFO(6..8)` and MUMPS in `RINFO(7..8)`).
+    This is the criterion with a principled stopping value: `t` on the order
+    of `eps` says "as good as a backward-stable factorization would have
+    returned," independent of `n` and of row scaling.
+- **What you get back.** The `*_refined_into` entry points now return a
+  `RefineOutcome { steps, relative_residual, stop }` instead of `()`, where
+  `stop` is `Converged | MaxSteps | Stagnated | Diverged`. All three values
+  are already computed by the loop, so reporting them costs nothing. The
+  full `RefinementDiagnostics` is deliberately *not* returned here: it
+  computes a Hager-Higham condition estimate costing 3-5 extra solves, which
+  would multiply exactly the cost this change exists to cut. Ask for it
+  explicitly via `solve_sparse_refined_with_diagnostics` when you want it.
+- **Migration.** `RefineOptions::default()` is unchanged and existing callers
+  compile unchanged — the new return value is not `#[must_use]`. Build a
+  criterion with `RefineOptions::with_target(t)`,
+  `RefineOptions::with_backward_error(t)`, or `.and_stop(...)` /
+  `.and_max_steps(...)` on an existing value.
+- **Note on scale.** `||r||_2 / ||b||_2` can look perfect while the solution
+  is not backward stable: on the worked badly-scaled 2x2 in the research
+  note it reads `1.0e-36` where the componentwise error is `5.0e-13`. Prefer
+  `BackwardError` unless you specifically want a normwise target.
+- **Oracle.** The backward-error formula is LAPACK `dgerfs`'s, including its
+  `safe1`/`safe2` underflow guard, with one documented deviation: a row whose
+  denominator is exactly zero contributes 0 rather than LAPACK's 1, because
+  such a row forces `r_i == 0` exactly. Expected values for the four unit
+  tests were computed independently in Python from the published formula
+  before any Rust was written.
+- See `dev/research/issue-190-refine-target.md` and
+  `dev/plans/issue-190-refine-target.md`.
+
+
 ### Changed - `needs_refinement` now measures growth against the pivot threshold, not a fixed 1e6
 
 - **What changed.** `Factors::needs_refinement` was set whenever any
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,29 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Changed - `needs_refinement` now measures growth against the pivot threshold, not a fixed 1e6
+
+- **What changed.** `Factors::needs_refinement` was set whenever any
+  `|L_ij| > 1e6`. That is a statement about the *magnitude* of `L`, which is
+  a property of the caller's scaling, not about growth, which is a property
+  of the factorization. It now compares against `1/u`, the multiplier bound
+  that threshold partial pivoting with `pivot_threshold = u` actually
+  promises. With no threshold in force (`u = 0`) the factorization makes no
+  such promise and the old absolute `1e6` remains as the fallback.
+- **Why it is the right bound.** TPP accepts a candidate only within a factor
+  `u` of its column maximum, so `|L_ij| <= 1/u` is the contract. Exceeding it
+  means the pivoting did not deliver its own guarantee on this matrix, which
+  is exactly when plain forward/back substitution is untrustworthy. The bound
+  also moves with the threshold, so `Solver::increase_quality` walking `u`
+  from 1e-8 towards `pivtol_max = 0.5` tightens the signal as it tightens the
+  promise.
+- **Who is affected.** Callers reading `factors.needs_refinement` on
+  badly-scaled systems will see far fewer spurious `true`s. A genuine
+  violation still flags: at `u = 1e-8` a `max|L|` of 4.2e13 is five orders
+  past the promise and sets the flag as before.
+- Covered by four new cases in `growth_flag_tests`, including that the same
+  `L` is accepted at `u = 1e-8` and rejected at `u = 0.5`.
+
 ### Fixed - multi-RHS solve was up to 1.6x slower when `nrhs` was not a multiple of 8
 
 - **What changed.** The row-major multi-RHS work buffers (`y`, `w`) used a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 All notable changes to FERAL will be documented in this file.
 
+## [Unreleased]
+
+### Changed - the benchmark harness now times the solve core a host actually gets (issue #189)
+
+- **What changed.** `cargo run --bin bench` measured its sparse solve column
+  with `solve_sparse_refined`, which pins `SolveCore::SharedVector`. Every
+  host reaching FERAL through `Solver` gets `SolveCore::Auto`, which picks
+  between the shared-vector and contribution-block cores per factor. The
+  harness now calls `solve_sparse_refined_auto_into`, so the published number
+  describes the code path hosts run.
+- **Why it matters if you read the numbers.** On seven large KKT matrices,
+  `Auto` is 1.30x faster than `SharedVector` (geomean, n = 8k-181k), and the
+  old harness could not see any of it. On the ~154k-matrix regression corpus
+  the aggregate `solve/MUMPS` geomean and p50 are unchanged at 0.08, because
+  that corpus is almost entirely small matrices where `Auto` selects
+  `SharedVector` anyway.
+- **The harness stays serial** (`parallel = false`), so the solve column
+  remains a single-threaded measurement comparable with previous releases;
+  the schedule choice is bit-neutral and measures 0.98-1.01x, so nothing is
+  lost by holding it fixed.
+- **No library behaviour changes.** This is the benchmark binary only.
+
 ## [0.17.0] - 2026-08-19
 
 ### Fixed — the tree-parallel solve no longer runs on trees too thin to pay for it (issue #175)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,35 @@ All notable changes to FERAL will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed - multi-RHS solve was up to 1.6x slower when `nrhs` was not a multiple of 8
+
+- **What changed.** The row-major multi-RHS work buffers (`y`, `w`) used a
+  leading dimension of exactly `nrhs`. When `nrhs` is not a multiple of 8,
+  every row of every supernode panel straddles cache lines, and the
+  misalignment compounds across the gather, the panel kernels, and the
+  scatter. The BLAS-3 path now pads the row stride up to a whole 8-column
+  tile, so every row starts on a 64-byte boundary. Separately, the panel
+  GEMM's `nrhs % 8` column tail ran an unblocked triple loop with no data
+  reuse; it now runs full-width register tiles with the unused lanes
+  zero-padded.
+- **Who is affected.** Anyone calling `solve_sparse_many`,
+  `solve_sparse_many_into`, or `solve_many_refined` with `nrhs >= 32` that is
+  not a multiple of 8 — 7 values in 8.
+- **Measured.** Geomean **1.36x** faster on the multi-RHS solve at
+  `nrhs = 33` across bcsstk38, r05_kkt and bratu3d (1.21x, 1.59x, 1.32x
+  anchored on `nrhs = 32`, whose code path is unchanged; raw times 1.22x,
+  1.22x, 1.32x). Before the fix, `nrhs = 31` cost **1.4-1.8x** `nrhs = 32` on
+  all seven large KKT matrices despite doing 3% less work; it is now ~1.0x.
+- **Results are bit-for-bit unchanged.** The padding columns carry a zero
+  right-hand side and are never read back, so every real column's arithmetic
+  and its order are untouched. Verified three ways: a new 800-shape
+  `to_bits()` equality test over every `nrhs % 8` and `m_dim % 4` residue; the
+  rank-1-vs-BLAS-3 difference unchanged at all 105 measured (matrix, `nrhs`)
+  points; and all 13 multi-RHS tests including the two
+  `assert_eq!(max_diff, 0.0)` band contracts. `BLAS3_NRHS_THRESHOLD` is
+  unchanged at 32.
+- See `dev/research/blas3-threshold-refit.md`.
+
 ### Changed - the benchmark harness now times the solve core a host actually gets (issue #189)
 
 - **What changed.** `cargo run --bin bench` measured its sparse solve column
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,36 @@ All notable changes to FERAL will be documented in this file.
   computes a Hager-Higham condition estimate costing 3-5 extra solves, which
   would multiply exactly the cost this change exists to cut. Ask for it
   explicitly via `solve_sparse_refined_with_diagnostics` when you want it.
-- **Migration.** `RefineOptions::default()` is unchanged and existing callers
-  compile unchanged — the new return value is not `#[must_use]`. Build a
-  criterion with `RefineOptions::with_target(t)`,
+- **Migration — BREAKING. An earlier version of this entry said the opposite
+  and was wrong.** It claimed "existing callers compile unchanged — the new
+  return value is not `#[must_use]`." That is false, and a downstream build
+  (pounce) disproved it. `#[must_use]` governs only whether *discarding* a
+  value warns at statement position; it does nothing for the value's type.
+  Three `Solver` methods now return `Result<RefineOutcome, FeralError>` where
+  they returned `Result<(), FeralError>`:
+  - `Solver::solve_refined_into`
+  - `Solver::solve_many_refined_into`
+  - the parallel refined entry point reached from `solve_refined_into`
+
+  Any caller that uses the result in *expression* position — a `match` arm, a
+  tail expression in a function returning `Result<(), _>`, a `let () =`
+  binding — fails to compile with `E0308: mismatched types`. Reported
+  downstream as incompatible `match` arms at `pounce-feral/src/lib.rs:1257`.
+
+  The fix at a call site that does not want the outcome is `.map(|_| ())`:
+
+  ```rust
+  // before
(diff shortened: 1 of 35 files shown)


jkitchin:
## Downstream measurement from pounce, plus an unlisted API break

I wired pounce to this branch by path (`feral = { path = ... }`) and measured
the stop-criterion choice on a real IPM workload. One binary throughout; the
only difference between arms is the `StopCriterion` handed to `RefineOptions`,
selected by a scratch env knob. Two rounds, round-robin, spread under 1.5%.

**`laptime`** — 58 014 variables, L-BFGS, 126 028-dimension KKT, `max_iter=60`
(all arms run the full 60 iterations):

| arm | stop criterion | Overall | Factor | Back-solve | vs A |
|---|---|---:|---:|---:|---:|
| A | `EpsSqrtN` — what 0.17.0 ships | 58.6 s | 9.27 s | 46.7 s | — |
| B | refinement **off** | 13.9 s | 5.85 s | 5.44 s | -76% |
| C | `RelativeResidual(1e-10)` | 22.8 s | 6.48 s | 13.77 s | -61% |
| D | **this PR's new default** | 69.4 s | 10.39 s | 56.4 s | **+18%** |

**`dirichlet120`** — exact Hessian, runs to convergence. All four arms reach
`Optimal` in 56 iterations at `3.5036085697750997e-02` (C differs in the last
digit). Back-solve 3.59 / 1.15 / 1.51 / 3.80 s for A/B/C/D. The model is
factorization-dominated, so it does not discriminate; recording it so the
`laptime` number is not read as the only case looked at.

### On the trade

Your table calls the cost a tail effect — solve/MUMPS p90 +33%, p99 +52%. On
`laptime` it is not the tail, it is the whole run: **+18% overall and +21% on
back-solve against 0.17.0**, every iteration. So the reviewer decision you asked
for has at least one downstream data point where the shipped default is a
straight regression, and I would rather you had it than not.

That said, it does not block pounce, for a reason specific to how pounce uses
feral: pounce's IPM already runs Ipopt's own refinement over the *unreduced*
augmented system, so we are turning the backend loop off by default anyway
(gh#698 observation 5). With `refine = false` nothing in this PR executes on
that path. The callers that keep `FeralConfig::default()` — pounce-convex,
pounce-rs, pounce-py — refine their own system without an escalation rung, and
those are the ones that genuinely want this fix, cost included.

Arm C is the part I would highlight as the win. `RelativeResidual` lets a caller
say what convergence means, and it recovers 61% of arm A's cost while keeping a
real accuracy criterion. It is also exactly the knob pounce's `FeralConfig::refine_target`
field documents itself as a placeholder for ("it goes away when feral#190
lands") — that pre-check solves unrefined, measures, and on a miss calls
`solve_refined`, which redoes the initial solve. `RelativeResidual` deletes both
the field and the wasted substitution. Thank you for that one.

Worth noting arm C still loses to arm B on this workload by 1.6x. That is not an
argument against the criterion; it is the nested-loop problem, and it is ours.

### API break not listed in the PR body

`Solver::solve_refined_into` and `solve_many_refined_into` now return
`Result<RefineOutcome, FeralError>` where 0.17.0 returned `Result<(), FeralError>`.
pounce does not compile against this branch until the call sites `.map(|_| ())`:

```
error[E0308]: `match` arms have incompatible types
    --> crates/pounce-feral/src/lib.rs:1257:29
     |
1253 |     (true, Some(m), true) => self.solver.solve_refined_into(m, rhs_vals, x_out, opts),
     |                              -------------------------------------------------------- this is found to be of type `Result<RefineOutcome, FeralError>`
```

Adding `stop` to `RefineOptions` is source-compatible for anyone using
`with_max_steps`, so the return type is the only thing that bites. Semver-major
for a 0.x crate — 0.18.0 — and worth a line in the PR body and the changelog so
downstreams are not surprised by it.

### Also

`BLAS3_NRHS_THRESHOLD` is still 32, so `a0b4d64`'s stride padding does not reach
pounce's multi-RHS back-solves: we cap `nrhs` at 16 deliberately, to stay under
that threshold and keep the blocked solve bit-identical to looping (pounce
gh#729). Not a request to change anything — just closing the loop that the
1.36x does not show up downstream for us.

jkitchin:
Both findings confirmed and acted on in `978b41e`. The API break is mine and the
changelog claimed the opposite of the truth.

## The break

Verified against the diff — it is three methods, not two:

- `Solver::solve_refined_into`
- `Solver::solve_many_refined_into`
- the parallel refined entry point reached from `solve_refined_into`

all `Result<(), FeralError>` → `Result<RefineOutcome, FeralError>`, introduced in
`f547bc5`.

The changelog said "existing callers compile unchanged — the new return value is
not `#[must_use]`." That is a non-sequitur and it is false. `#[must_use]` governs
whether *discarding* a value warns at statement position; it does nothing for the
value's type, so every expression-position use — your `match` arm, a tail
expression in a `-> Result<(), _>` function, a `let () =` — breaks. The entry now
names the three methods, shows the `.map(|_| ())` migration, and states 0.18.0
rather than 0.17.1. The PR body leads with it.

The actual six-file bump I left out of this PR deliberately: it goes through
`scripts/release-checklist.sh bump 0.18.0`, which keeps the root crate, both
lockfiles, and the Python trio in sync, and it pairs with renaming the changelog's
`[Unreleased]` heading. Doing half of it here would be worse than not doing it.
Say the word if you want it in this PR anyway.

## The laptime number

This is the more useful of the two, and it is worse than what I reported. My
corpus A/B said "tail regression, median untouched" — p90 +33%, p99 +52%. Your
`laptime` arm D is +18% overall and +21% on back-solve, on every one of 60
iterations. That is not the tail; the corpus percentiles simply do not have a
128k-dimension repeatedly-refined IPM in them, so they described the shape of the
cost and understated its incidence. Both tables are now in the changelog, with
`dirichlet120` recorded alongside so `laptime` does not read as the only case
looked at.

Arm B at -76% is the number I would not have predicted. Refinement off beating
`RelativeResidual(1e-10)` by 1.6x on a workload that still converges to the same
objective says the backend loop is redundant work under your outer loop, which is
your nested-loop point, and I agree it is not an argument against the criterion.

## On the default

I am leaving it as-is rather than flipping it, because you said the callers that
keep `FeralConfig::default()` — pounce-convex, pounce-rs, pounce-py — are the ones
that want this fix with the cost. That is the decision I did not want to make
unilaterally, and it is now made on your data rather than my inference. If you
would rather ship `EpsSqrtN` as the default and have every caller opt in, it is
one line at `src/numeric/solve.rs:2140` and I will redo the corpus A/B against it.

Noted on `BLAS3_NRHS_THRESHOLD`: capping `nrhs` at 16 to stay under 32 means
`a0b4d64`'s 1.36x is downstream-invisible for pounce. I have said so explicitly in
the PR body so the gain is not read as something you are getting.

jkitchin:
Downstream (pounce) confirmation, since #190 was filed from here.

## This closes both asks

`StopCriterion::RelativeResidual(f64)` on `RefineOptions` is ask 1, and
`RefineOutcome` out of `solve_refined_into` / `solve_many_refined_into` is
ask 2. Nothing further is needed from our side.

## The `refine_target` retirement is worth more than it reads

`FeralConfig::refine_target` is not just a redundant knob — the block it
guards (`crates/pounce-feral/src/lib.rs:1189-1224`) solves once *unrefined*,
measures `‖b − Ax‖₂/‖b‖₂` itself, and on a miss falls through to the refined
entry point, **which redoes that initial solve internally**. So today every
back-solve that needs refinement pays for one extra substitution pass to
discover it. All of that deletes when `stop` lands. Its option
(`feral_refine_target`) and env fallback go with it.

## Don't hold the PR on our account over the default

Since `stop` is per call, we will set it explicitly and never see
`RefineOptions::default()`. Whether the conjunctive default is right for
other callers is genuinely your decision, not ours, and the +18% we
contributed to the table shouldn't be read as a veto.

For the record on the lever itself, from the same table:
`RelativeResidual(1e-10)` takes `laptime` from 58.6 s to 22.8 s overall and
46.7 s to 13.8 s on back-solve — **−61%**, on a 126,028-dimension KKT over
60 IPM iterations. That is the number that closes #190 for us.

## The componentwise finding is probably diagnosing one of *our* defects

Worth saying explicitly, because it is easy to read this PR as purely a
feral-internal accuracy story. The fixtures where the normwise rule leaves
`ω > 10·√ε` — `DEGENLPB_0045/46/47`, `DEGENLPA_0065`, `ACOPP14_0003` — are
degenerate-LP and ACOPF KKTs, which is exactly the shape our corpus is made
of. And #190's own mechanism section argued that the step cap behaves like a
lottery because "smallest `‖r‖₂` on the condensed system has no reliable
relationship to which one gives the better Newton step". Componentwise
backward error is a much better proxy for that than the normwise residual is.

Two open pounce results we intend to re-run under `BackwardError` and the
conjunction once this lands, and report back here:

- `cresc4` at `max_steps=1` returns a **false `InfeasibleProblemDetected` on
  a feasible problem** in 32 iterations. Not a slower route to the right
  answer — a wrong answer.
- the gh#590 badly-scaled LP (data scale 1e11), which is the one fixture that
  `feral_refine_steps=0` loses outright.

If either is the componentwise defect, the right pounce configuration is
probably not one criterion at all: `RelativeResidual` on the hot path,
the conjunction where accuracy is actually failing. Per-call `stop` makes
that expressible, which is the part of this design we care most about.

Happy to supply either KKT system if it is useful on the feral side.

## Cost of the break to us: five lines

All four refined/unrefined call sites feed one `match` at
`crates/pounce-feral/src/lib.rs:1233`, so `Ok(())` becomes `Ok(outcome)` and
we start consuming the outcome rather than discarding it. Not a blocker, and
we would rather have the value than not.

## Sequencing

This and #193 both touch `src/numeric/solver.rs`, and both are 0.18.0
material. If they can land together we do one pin bump and one downstream PR
covering the `refine_target` retirement, the `reset_quality` plumbing, and
the pounce gh#857 re-baselining. Three separate feral releases would cost us
three sweeps.

## One question

Does `RefineOutcome` carry the achieved residual / `ω` as well as the step
count? Ask 2 was phrased as "see whether the budget bound", and the step
count alone answers that — but if the achieved value is already in there we
would log it per back-solve, which would make the next round of this
conversation a measurement instead of an inference.

jkitchin:
> Does `RefineOutcome` carry the achieved residual / ω as well as the step count?

It carries the residual, and as of the merged version it carries ω too — it didn't before you asked, so thanks for asking.

`RefineOutcome` is now:

```rust
pub struct RefineOutcome {
    pub steps: usize,
    pub relative_residual: f64,
    pub backward_error: f64,   // new
    pub stop: RefineStop,
}
```

`backward_error` is the achieved componentwise ω of the *returned* iterate (the best one seen, same iterate `relative_residual` describes), and for a multi-RHS call it is the worst column — matching how `steps` and `relative_residual` aggregate.

The reason to add it now rather than in 0.19.0: ω was already computed at every construction site and thrown away, `RefineOutcome` is new and unreleased so the field costs no additional break, and 0.18.0 makes ω *the quantity the default criterion certifies*. Getting `Converged` back without being able to see the certificate — or having to spend a second `abs_symv` recomputing it — was the wrong shape.

**One sharp edge worth knowing about before you consume it.** ω is only formed when it is the criterion; under `EpsSqrtN` and `RelativeResidual`, and under `max_steps = 0`, the loop never computes it. On those paths the field is **`NaN`, meaning *not measured***. That is deliberately distinct from `INFINITY`, which is a real measurement — a non-finite iterate. So:

```rust
if !o.backward_error.is_nan() && o.backward_error > my_bound { ... }
```

and not a bare `o.backward_error > my_bound`, since `NaN` fails every comparison and would silently read as "fine". The default criterion does form ω, so on the default path you always get a number.

Two other things from this round that touch you:

- **A real bug in `backward_error` itself, fixed here.** On a non-finite iterate every row's `|r|/d` was `NaN`, `NaN` loses every `>` comparison, so the running max was never updated and the function returned `ω = 0.0` — a diverged solve certifying itself as *exactly* backward stable. Under `StopCriterion::BackwardError(t)`, which this PR is what first exposes publicly, that stops refinement on its first test and returns `Converged` with a garbage iterate. It now returns `INFINITY`. `EpsSqrtN` never had the hole, and the shipped default conjunction is covered by its normwise half, so the exposure was the bare `BackwardError` option you'd be opting into.
- **Two more breaks are now in the CHANGELOG** that weren't: `RefineOptions { max_steps: k }` struct literals are `E0063` (the new `stop` field), and the derive dropped `Eq` (`StopCriterion` carries an `f64`), so `RefineOptions` as a `HashMap`/`HashSet` key or under a `T: Eq` bound is `E0277`.

PR #193, opened 2026-08-29
feat: add Solver::reset_quality() to bound escalation lifetime (#192)

### LINKED ISSUE #192: increase_quality escalation is permanent and unscopable: quality_level only ratchets up, with no reset
## Summary

`Solver::increase_quality` escalates permanently and cannot be reverted or
scoped: `quality_level` only ratchets `Baseline → ScalingEnabled → PivotRaised
→ Exhausted`, and there is no reset. So an escalation chosen for *one* hard
factorization governs **every** factorization for the rest of the solve.

That is fine for an escalation that is monotone in robustness. FERAL's is not
monotone in *trajectory*, and downstream that costs whole solves.

## Why the persistence is load-bearing

Ipopt's `IncreaseQuality` contract (`IpPDFullSpaceSolver.cpp:296`) is "this
factorization cannot deliver — escalate and refactor". MA57 answers it by
raising `pivtol` toward `pivtolmax`
(`IpMa57TSolverInterface.cpp:832`): strictly more conservative each time, so
keeping it raised for the remainder of the solve can only make the
factorization safer, and never reverting is harmless.

FERAL's ladder (`Identity → InfNorm`, then `pivot_threshold^0.75`) changes
**which pivots are taken**. That is a lateral move, not a monotone one — the
factorization is different, not uniformly better — and because it persists, the
whole remaining trajectory is different too, including any restoration
sub-problem's factorizations.

## Measured downstream (pounce gh#850)

On `square_flowsheet_resto` (a square flowsheet NLP), wiring
`increase_quality` through cost both solve arms:

| | with escalation | without |
|---|---|---|
| exact Hessian | `RestorationFailed`, 131 iters | **`Optimal`, 99** |
| limited-memory | 3000 iters, at the cap | **`Optimal`, 178** |

The rung fires exactly **twice** there: once in the main solve at iteration 25,
once inside the restoration sub-solve at `76r`.

Three things rule out the obvious workarounds, all measured with a
process-global firing cap:

1. **Scoping it out of restoration would not help.** Allowing *only* the first
   (main-solve) firing still loses the leg.
2. **A firing count does not separate good from bad.** `deb7` and
   `square_flowsheet_resto` each fire it exactly twice on their exact legs —
   one gains 16% of its iterations, the other loses its verdict.
3. **Simply not escalating is not the answer either.** The escalation is
   load-bearing elsewhere: a 12-variable watchdog model ends
   `Solved_To_Acceptable_Level` at `obj = 3.7e-6` with it and at `obj = 3.42`
   against `f* = 0` without it, and it buys 15–25% of the iterations on five
   other models.

So the escalation is genuinely useful and genuinely destructive, and the
distinguishing factor is not *whether* to escalate but *for how long*.

## What would fix it

A way to bound the escalation's lifetime rather than have it own the solve.
Any of these would do; listing them as options rather than a preference:

- `Solver::reset_quality()` (or `set_quality_level(Baseline)`), so a caller can
  restore the factory parameters once the system that needed the escalation is
  past.
- A scoped/guard form — escalate for the next `factor()` only, then fall back.
- Reporting enough for a caller to do it itself: `quality_level()` is already
  public, so a paired setter is the smallest possible change.

The smallest useful increment is the reset. Downstream can then re-baseline at
whatever boundary makes sense (a new major iteration, entering restoration)
without FERAL having to know anything about the caller's algorithm.

## Repro

`increase_quality` is reached through `pounce`'s `FeralSolverInterface`. In
pounce at `84ad7da1`, `feral_increase_quality=no` declines the rung and both
legs above solve; the default leaves it on. `pounce-cli`'s
`issue850_increase_quality_regression.rs` pins the trade in both directions,
and `dev-notes/second-opinion-promotions-in-the-sweep.md` carries the full
measurement.

Filed from the pounce side; happy to supply the KKT systems if that helps.

jkitchin:
Closes #192.

## What

New `Solver::reset_quality() -> bool`, plus its Python binding. It reverts every escalation `increase_quality` applied — the `Identity → InfNorm` scaling flip and the raised `bk.pivot_threshold` — restoring the parameters the caller configured and returning `quality_level()` to `Baseline`. Returns `true` if an escalation was undone, `false` if already at baseline. Valid from any level, `Exhausted` included.

## Why

`increase_quality` was a one-way ratchet with no way back, so an escalation chosen for *one* hard factorization governed **every** later factorization for the life of the `Solver`.

Ipopt exposes no reset because MA57 answers `IncreaseQuality` by raising `pivtol` toward `pivtolmax` (`IpMa57TSolverInterface.cpp:832`) — strictly more conservative each time. That is monotone in *robustness*, so leaving it raised for the rest of a solve can only make later factorizations safer, and never reverting is harmless. FERAL's rungs change *which pivots are taken*: the factorization is different, not uniformly better, and persisting it changes the caller's whole remaining trajectory, including any restoration sub-problem's factorizations.

`dev/research/pounce-integration-interface.md` had deferred this — "FERAL should match [Ipopt] until there is evidence to add it." Issue #192 is that evidence; this takes up the deferral rather than reversing a decision.

### Downstream evidence (pounce gh#850, from the issue)

On `square_flowsheet_resto` the rung fires exactly twice and costs both solve arms:

| leg | with escalation | without |
|---|---|---|
| exact Hessian | `RestorationFailed`, 131 iters | `Optimal`, 99 |
| limited-memory | 3000 iters, at the cap | `Optimal`, 178 |

Not escalating is not the fix either: a 12-variable watchdog model ends `Solved_To_Acceptable_Level` at `obj = 3.7e-6` with the escalation and at `obj = 3.42` against `f* = 0` without it, and the rung buys 15–25% of the iterations on five other models. The distinguishing variable is how long the escalation lasts, not whether it fires.

## Design notes

**The baseline is snapshotted lazily**, on the transition out of `Baseline`, not at construction. The `with_*` builders are consuming and run *after* `with_params`, so a construction-time snapshot would record pre-builder values and a reset would silently discard the caller's configuration. It also makes the round trip exact by construction: `reset` → `increase` retraces the same rungs a freshly constructed `Solver` would.

**The reset touches nothing else**, mirroring what `increase_quality` leaves alone — not `last_symbolic`, not `mc64_scaling_cache`, not the independent latches. The cached symbolic factorization is scaling-invariant since scaling moved to the numeric phase, so re-baselining costs no re-analysis, exactly as escalating costs none.

**The scoped-guard form the issue also lists was considered and not implemented.** A guard fixes the scope at one `factor()`, but the Ipopt contract loop escalates repeatedly until the factorization delivers, so the natural scope is the caller's own retry loop — a shape FERAL does not know and should not assume. It stays expressible in terms of the reset if evidence for it later arrives. Recorded as considered-and-not-taken in the research note, not reframed as a rejection.

## Evidence

Six unit tests (R1–R6) and integration test I9, all written RED first (`error[E0599]: no method named reset_quality`, 6 compile errors), plus one Python test.

Every oracle is a parameter value the *caller* supplied, never implementation output:
- **R5** constructs with `pivot_threshold = 0.25` and fails if the reset restores `NumericParams::default()`.
- **R4**'s oracle is the ladder traversal recorded *before* any reset exists in the state.
- **R6** pins the lazy snapshot against a construction-time one.
- **I9** pins that re-baselining costs no symbolic re-analysis (`symbolic_call_count` never moves off 1).

```
cargo test           -> 448 passed, 0 failed, 7 ignored (lib);
                        all integration suites green
pytest tests/        -> 16 passed (built via maturin develop, run locally)
cargo clippy --all-targets -- -D warnings -> clean
cargo fmt --check    -> clean
```

The escalation ladder itself is untouched — rungs, the `0.75` exponent, `pivtol_max` — and U1–U5 pass unchanged, so a caller that never calls `reset_quality` sees byte-identical behaviour.

## What was not measured

**The exit partition was not measured on this branch.** The 153k-matrix corpus is not present in the session container, so both partitions report `N/A` and there is no comparison against 2026-08-19-05's 1.58 / 1.58 to report — favourable or otherwise. What did run is clean (inertia 2/2 vs MUMPS, residual 2/2, worst 1.26e-16 on the 2 synthetic KKT matrices the bench generates).

The diff has no mechanism by which it could regress performance — a pure API addition whose new code path runs only when a caller calls `reset_quality()`, with no numeric kernel, ordering, scaling, or pivoting code touched. That is an argument from the diff's shape, not a number, and it is worth confirming on a host with corpus access before release.

## Also in this PR

- `CHANGELOG.md` Unreleased entry; `FERAL-PROJECT-SPEC.md` POUNCE API listing.
- Research note `dev/research/issue-192-quality-reset.md`, plan `dev/plans/issue-192-quality-reset.md`, session checkpoint `dev/sessions/2026-08-29-01.md`, journal, and two appended entries in `dev/decisions.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SmY6FDHgsdqSMTi6GsqAz7)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,16 @@ All notable changes to FERAL will be documented in this file.
   binding `Solver.reset_quality()`). It reverts every escalation
   `increase_quality` applied — the `Identity → InfNorm` scaling flip and the
   raised `bk.pivot_threshold` — restoring the parameters the caller had
-  configured and returning `quality_level()` to `Baseline`. Returns `true` if
-  an escalation was undone, `false` if the solver was already at baseline.
-  Valid from any level, `Exhausted` included.
+  configured and returning `quality_level()` to `Baseline`. Valid from any
+  level, `Exhausted` included. Returns `true` only if a parameter was actually
+  restored to a different value — `false` both when the solver was already at
+  baseline and when the escalation itself moved nothing (a rung can fire, and
+  report `true`, without changing a parameter: from non-`Identity` scaling with
+  `pivot_threshold` already at `pivtol_max`, `0.5^0.75` clamps back to `0.5`).
+  Either way the level is re-baselined and the ladder is armed again.
+  `with_scaling` applied while an escalation is live re-states the caller's
+  baseline, so a later reset restores the strategy pinned last rather than
+  silently undoing it (the issue-#51 escape hatch keeps working mid-solve).
 - **Why.** `increase_quality` was a one-way ratchet with no way back, so an
   escalation chosen for *one* hard factorization governed **every** later
   factorization for the life of the `Solver`. Ipopt exposes no reset because
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,42 @@
 
 All notable changes to FERAL will be documented in this file.
 
+## [Unreleased]
+
+### Added — `Solver::reset_quality()` bounds the lifetime of a quality escalation (issue #192)
+
+- **What changed.** New `Solver::reset_quality() -> bool` (and the Python
+  binding `Solver.reset_quality()`). It reverts every escalation
+  `increase_quality` applied — the `Identity → InfNorm` scaling flip and the
+  raised `bk.pivot_threshold` — restoring the parameters the caller had
+  configured and returning `quality_level()` to `Baseline`. Returns `true` if
+  an escalation was undone, `false` if the solver was already at baseline.
+  Valid from any level, `Exhausted` included.
+- **Why.** `increase_quality` was a one-way ratchet with no way back, so an
+  escalation chosen for *one* hard factorization governed **every** later
+  factorization for the life of the `Solver`. Ipopt exposes no reset because
+  MA57 answers `IncreaseQuality` by raising `pivtol` toward `pivtolmax` —
+  monotone in robustness, so leaving it raised can only make later
+  factorizations safer. FERAL's rungs change *which pivots are taken*: the
+  factorization is different, not uniformly better, and persisting it changes
+  the caller's whole remaining trajectory.
+- **Evidence (pounce gh#850).** On `square_flowsheet_resto` the rung fires
+  twice and costs both solve arms: the exact-Hessian leg goes from `Optimal`
+  in 99 iterations to `RestorationFailed` in 131, and the limited-memory leg
+  from `Optimal` in 178 to 3000 iterations at the cap. Declining to escalate
+  is not the fix either — a 12-variable watchdog model ends at `obj = 3.7e-6`
+  with the escalation and at `obj = 3.42` against `f* = 0` without it, and the
+  rung buys 15–25% of the iterations on five other models. The distinguishing
+  variable is how long the escalation lasts, not whether it fires.
+- **Nothing else changes.** The escalation ladder — its rungs, the `0.75`
+  exponent, `pivtol_max` — is untouched, and a solver that never calls
+  `reset_quality` behaves exactly as before. The reset touches only the two
+  escalated parameters and the level: the cached symbolic factorization
+  survives (it is scaling-invariant since scaling moved to the numeric phase),
+  so re-baselining costs no re-analysis, exactly as escalating costs none.
+  `reset_quality` followed by `increase_quality` retraces the same rungs a
+  freshly constructed `Solver` would.
+
 ## [0.17.0] - 2026-08-19
 
 ### Fixed — the tree-parallel solve no longer runs on trees too thin to pay for it (issue #175)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,14 +36,26 @@ All notable changes to FERAL will be documented in this file.
   with the escalation and at `obj = 3.42` against `f* = 0` without it, and the
   rung buys 15–25% of the iterations on five other models. The distinguishing
   variable is how long the escalation lasts, not whether it fires.
-- **Nothing else changes.** The escalation ladder — its rungs, the `0.75`
-  exponent, `pivtol_max` — is untouched, and a solver that never calls
-  `reset_quality` behaves exactly as before. The reset touches only the two
-  escalated parameters and the level: the cached symbolic factorization
-  survives (it is scaling-invariant since scaling moved to the numeric phase),
-  so re-baselining costs no re-analysis, exactly as escalating costs none.
-  `reset_quality` followed by `increase_quality` retraces the same rungs a
-  freshly constructed `Solver` would.
+- **Guaranteed, not incidental.** Three properties are contract, each pinned
+  by a named test; a change that breaks one is a breaking change, not a detail.
+  1. *A caller that never calls `reset_quality` sees byte-identical behaviour
+     to 0.17.0.* The escalation ladder is untouched — its rungs, the `0.75`
+     exponent, `pivtol_max` — and the snapshot is inert state. Pinned by U1–U5,
+     which predate the method and pass unchanged. Downstream release processes
+     that diff a fixture sweep against a baseline binary depend on this not
+     drifting silently.
+  2. *`reset_quality` then `increase_quality` retraces exactly the rungs a
+     freshly constructed `Solver` would.* Pinned by R4 (records the traversal
+     before any reset exists, asserts the second equals it) and R6.
+  3. *Re-baselining costs no symbolic re-analysis.* The cached
+     `SymbolicFactorization` survives, exactly as it survives an escalation.
+     A cost guarantee: a caller may reset once per IPM iteration, so a reset
+     that forced re-analysis would be a silent per-iteration cost. Pinned by
+     integration test I9 via `symbolic_call_count`.
+- **Nothing else changes.** The reset touches only the two escalated
+  parameters and the level — not `last_factors`, not `mc64_scaling_cache`, and
+  none of the independent latches — mirroring what `increase_quality` leaves
+  alone.
 
 ## [0.17.0] - 2026-08-19
 
--- a/FERAL-PROJECT-SPEC.md
+++ b/FERAL-PROJECT-SPEC.md
@@ -564,7 +564,16 @@ pub enum FactorStatus {
 /// Stage 1: activate enhanced scaling (if not already on).
 /// Stage 2: raise pivot threshold parameter u.
 /// Modeled on Ipopt's IncreaseQuality() — see IpTSymLinearSolver.cpp:432-441.
+/// Persists across factorizations until reset_quality() reverts it.
 pub fn increase_quality(&mut self) -> bool;
+
+/// Revert every escalation applied by increase_quality(), restoring the
+/// configured scaling strategy and pivot threshold and returning to
+/// QualityLevel::Baseline. Returns true if an escalation was undone.
+/// Ipopt has no counterpart: MA57's escalation is monotone in robustness,
+/// FERAL's changes which pivots are taken, so its lifetime has to be
+/// boundable by the caller (issue #192).
+pub fn reset_quality(&mut self) -> bool;
 
 /// Number of negative eigenvalues from last factorization.
 pub fn num_negative_eigenvalues(&self) -> usize;
--- a/crates/feral-diagnostics/src/bin/diag_acopr14.rs
+++ b/crates/feral-diagnostics/src/bin/diag_acopr14.rs
@@ -181,12 +181,10 @@ fn run_one(matrix_name: &str) {
     let n_schur = schur.len();
     let bytes = std::fs::read(&bin_path).unwrap();
     let mumps: Vec<f64> = bytes
-        .chunks_exact(8)
-        .map(|c| {
-            let mut b = [0u8; 8];
-            b.copy_from_slice(c);
-            f64::from_le_bytes(b)
-        })
+        .as_chunks::<8>()
+        .0
+        .iter()
+        .map(|c| f64::from_le_bytes(*c))
         .collect();
 
     println!("  n={} n_schur={}", n, n_schur);
--- a/crates/feral-diagnostics/src/bin/diag_schur_parity.rs
+++ b/crates/feral-diagnostics/src/bin/diag_schur_parity.rs
@@ -113,10 +113,8 @@ fn read_mumps_schur(json_path: &Path) -> Option<MumpsSchurRef> {
         return None;
     }
     let mut data = Vec::with_capacity(n_schur * n_schur);
-    for chunk in bytes.chunks_exact(8) {
-        let mut buf = [0u8; 8];
-        buf.copy_from_slice(chunk);
-        data.push(f64::from_le_bytes(buf));
+    for chunk in bytes.as_chunks::<8>().0 {
+        data.push(f64::from_le_bytes(*chunk));
     }
     Some(MumpsSchurRef {
         n,
@@ -138,10 +136,8 @@ fn read_dense_oracle(mtx_path: &Path, n_schur: usize) -> Option<Vec<f64>> {
         return None;
     }
     let mut data = Vec::with_capacity(n_schur * n_schur);
-    for chunk in bytes.chunks_exact(8) {
-        let mut buf = [0u8; 8];
-        buf.copy_from_slice(chunk);
-        data.push(f64::from_le_bytes(buf));
+    for chunk in bytes.as_chunks::<8>().0 {
+        data.push(f64::from_le_bytes(*chunk));
     }
     Some(data)
 }
--- a/crates/feral-ordering-core/src/quotient_graph/algo.rs
+++ b/crates/feral-ordering-core/src/quotient_graph/algo.rs
@@ -1204,9 +1204,8 @@ pub fn finalize_step_amf(
             ws.nv[i] = nvi;
             let degme_i = degme_i32;
             let nvi_i = nvi;
-            let rmf: f64;
             let deg_i = ws.degree[i];
-            if (deg_i as usize) + (degme_i as usize) > nleft {
+            let rmf: f64 = if (deg_i as usize) + (degme_i as usize) > nleft {
                 // Saturated branch. RMF1 uses original DEG.
                 let deg_f = deg_i as f64;
                 let rmf1 = deg_f * (deg_f - 1.0 + 2.0 * degme_i as f64) - ws.wf[i] as f64;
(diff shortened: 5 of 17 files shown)


jkitchin:
Downstream (pounce) confirmation, since #192 was filed from here.

## The API is sufficient — we are not asking for the guard form

`reset_quality()` plus the already-public `quality_level()` is expressively
complete for every lifetime we could want: per-matrix, per-major-iteration,
per-solve, and the scoped-guard shape itself (escalate, factor, reset). The
PR's reasoning for declining the guard — that the natural scope is the
caller's own retry loop, which feral should not have to model — is right, and
it is our retry loop, so we would rather own it.

## Where we would call it, and why it is the boundary

`PdFullSpaceSolver` already clears its own `augsys_improved` latch in two
places — `crates/pounce-algorithm/src/kkt/pd_full_space_solver.rs:445`
(`invalidate_aug_cache`) and `:647` (the dependency-tag cache miss, mirroring
upstream's `dummy_cache_` lookup). Both mirror Ipopt's `augsys_improved_`.

So pounce **already believes the escalation ends when the matrix changes**.
The backend simply never got the message, which is #192 restated from our
side. Those two lines are where the reset goes, and that placement answers
"for how long" with "exactly as long as the matrix it was chosen for".

Worth flagging against #192's own evidence: this is **not** what the
process-global firing cap measured. Capping at one firing still left that
escalation in force for the whole remaining solve — which is why "allowing
only the first firing still loses the leg". A reset at the matrix boundary
expresses something the cap could not, so that measurement does not predict
this outcome in either direction. We will re-run the corpus rather than
assume.

## Three contract asks — wording, not code

Everything below is asserted in the PR body and pinned by the tests. The ask
is only that it move into the doc comment / CHANGELOG, so it is contract
rather than implementation detail.

1. **"A caller that never calls `reset_quality` sees byte-identical
   behaviour."** Our release process diffs a 158-leg fixture sweep against a
   baseline binary and treats any moved line as a trajectory change requiring
   explanation. If that property is incidental, a later refactor moves our
   baseline silently, which is the exact failure mode our own post-mortems are
   about.
2. **"`reset` → `increase` retraces the same rungs a freshly constructed
   `Solver` would."** R4/R6 pin it. Same reason: our sweep compares against a
   binary, not against a spec, so a non-idempotent round trip shows up as
   unexplained corpus drift months later.
3. **"Re-baselining costs no symbolic re-analysis."** I9 pins it via
   `symbolic_call_count`. This one has a cost consequence rather than a
   correctness one: we would call `reset_quality()` on every dependency-tag
   cache miss, i.e. up to once per IPM iteration on some models. If that ever
   started triggering a re-analysis it becomes a per-iteration cost for us,
   silently.

## On the unmeasured exit partition

Low risk for us — a pure API addition whose new path runs only when called,
and we agree with the diff-shape argument. But we pin to a crates.io release
rather than a rev, so the corpus run wants doing before the **release**, not
necessarily before the merge.

## Sequencing

This and #191 both touch `src/numeric/solver.rs`, and both are 0.18.0
material. Landing them together means one pin bump here and one downstream
PR covering the `refine_target` retirement, this plumbing, and the pounce
gh#857 re-baselining. Separate releases would cost us a separate fixture
sweep each.

jkitchin:
## Why there are unrelated files in this diff

The `check` job was red, and **not because of this PR's feature work**. `git diff --stat origin/main...HEAD -- crates/` is empty — this branch touches no file under `crates/`, and the failure was `clippy::needless_late_init` at `crates/feral-ordering-core/src/quotient_graph/algo.rs:1207`.

The cause is `dtolnay/rust-toolchain@stable` now resolving to **1.98.0** (`88d9e12ae`, 2026-08-18), which added two new lints. My dev container was pinned at 1.94.1 (`e408947bf`, 2026-03-25), which is why local clippy *and* the pre-commit hook both passed while CI failed. `rustup update stable` reproduced the failure exactly. **Any PR opened against `main` today hits this**, so the fix is ported here rather than waited on; it no-ops once `main` carries its own.

Fixed in 9c5dfac, split by risk:

| sites | lint | action |
|---|---|---|
| `feral-ordering-core/.../algo.rs:1207` | `needless_late_init` | real fix — clippy's own suggested transformation |
| `diag_schur_parity.rs` (×2), `diag_acopr14.rs` | `chunks_exact_to_as_chunks` | real fix — `as_chunks::<8>()`, drops a `copy_from_slice` each |
| `src/dense/schur_kernel.rs` (×10) | `chunks_exact_to_as_chunks` | **file-scoped `#[allow]`, not a rewrite** |

### Why the kernel gets an allow

Those 10 sites are the 4-way unrolled SIMD bodies of `axpy_minus` / `axpy2_minus` — the hottest loop in the factorization, whose unroll depth and `into_remainder()` cleanup are a measured design (`dev/research/dense-kernel-*.md`). `as_chunks::<4>()` looks semantically identical and may well be neutral or better for codegen, but that's a hypothesis, and this container has no corpus to benchmark it against (the exit partition reports `N/A` here). Landing an unmeasured edit to that loop to satisfy a style lint inverts the project's order of operations.

Recorded in `dev/tried-and-rejected.md` as **deferred pending measurement, not a verdict** — a session with corpus access should sweep it and either land the rewrite with numbers or record the regression. Happy to drop the allow and do the rewrite instead if you'd rather.

### One process note

Fixing the ordering-core error revealed 12 further sites that cargo had never reached, because the dependency crate aborted the build first — CI had only ever reported the first error. Pushing the one-line fix the log named would have gone red again on lints that were visible locally. Hence updating the toolchain to match CI and re-running the whole workspace before pushing.

Validation, all on 1.98.0 matching CI:

```
cargo clippy --workspace --all-targets -- -D warnings -> clean
cargo fmt --all --check                               -> clean
cargo test --workspace -> total passed: 1196, total failed: 0
```

The 1196 includes the `feral-ordering-core` suite, which is what would catch the AMF refactor had it changed the ordering.

The feature commits (0952f99, f925589) are untouched by this.


---

jkitchin:
## Review

The core feature is sound. I checked the state machine rather than the diff alone: `quality_baseline` is written in exactly one place (the lazy snapshot on leaving `Baseline`) and cleared in exactly one (`reset_quality`'s `.take()`), `quality_level` is mutated nowhere else in `src/`, and `factor()` never writes `self.numeric_params.scaling` — it clones into `effective_params`. So the sticky-`Auto` pick and the MC64 scaling cache are genuinely unaffected, because stage 1 only fires on an explicit `Identity`. `increase_quality` from `Baseline` always returns `true` and always leaves `Baseline`, so `quality_baseline` is `Some` only after a real escalation and `reset_quality()` cannot report an undo that never happened via that route. R1–R6 and I9 pass.

The `quotient_graph/algo.rs` `let rmf` refactor is behaviour-preserving — both `ws.degree[i]` reads are ordered correctly around the mutation, with `deg_i` captured before it.

Three findings.

### 1. The `chunks_exact_to_as_chunks` allow blocks local commits on any toolchain below 1.98 — high

`src/dense/schur_kernel.rs:41` carries `#![allow(clippy::chunks_exact_to_as_chunks)]`. That lint does not exist before stable 1.98. On the clippy installed here:

```
$ clippy-driver -W help | grep -c "clippy::"
819
$ clippy-driver -W help | grep chunks_exact_to_as_chunks
(no match)
$ cargo clippy --lib -- -D warnings
error: unknown lint: `clippy::chunks_exact_to_as_chunks`
  --> src/dense/schur_kernel.rs:41:10
   = note: `-D unknown-lints` implied by `-D warnings`
```

`cargo clippy -- -D warnings` is exactly what the mandatory pre-commit hook runs, so this blocks **every** local commit, on any change, for anyone not yet on 1.98 — not just commits touching this file. CI uses `dtolnay/rust-toolchain@stable` and will be green, so CI will not catch it, and nothing in `Cargo.toml` declares a `rust-version` floor. The PR body's "`cargo clippy --all-targets -- -D warnings` -> clean" does not hold on a 1.96 toolchain.

Fix is one line — `#![allow(unknown_lints)]` immediately above it — which restores a clean `cargo clippy --all-targets -- -D warnings` on 1.96 while keeping the 1.98 suppression working.

### 2. `reset_quality` can silently discard configuration applied after escalating — medium

`src/numeric/solver.rs:1998`. `with_scaling` (line 798) is the only external writer of `numeric_params.scaling` and it does not touch `quality_baseline`:

```rust
Solver::with_params(np /* Identity */, sn)
    .increase_quality()                  // snapshots Identity, flips to InfNorm
    .with_scaling(Mc64Symmetric)         // the documented #51 escape hatch
    .reset_quality()                     // restores the stale Identity snapshot
```

The solver then factors with `Identity`, not the `Mc64Symmetric` the caller just pinned. This is the same failure mode the lazy-snapshot design was chosen to avoid, relocated: R6 covers the *pre*-escalation window, but the post-escalation window is unguarded. Either have `with_scaling` refresh or clear `quality_baseline`, or document reconfiguration-during-escalation as undefined.

### 3. `reset_quality() == true` can report an undo when no parameter ever moved — low

Same site. `increase_quality()` snapshots unconditionally on leaving `Baseline`. For a solver built with non-`Identity` scaling whose `bk.pivot_threshold` is already at `pivtol_max = 0.5`, `bump_pivot_threshold` computes `0.5^0.75 = 0.594`, clamps back to `0.5` — unchanged — and jumps straight to `Exhausted`. A caller keying retry telemetry on `reset_quality() == true` ("an escalation was undone", per the doc comment and R2) gets a false positive. Guard by snapshotting, or returning `true`, only when a parameter actually differs.

### Scope note

`9c5dfac` ("satisfy clippy 1.98's new lints") is unrelated to #192, touches ordering-core and the diagnostics binaries, and is the sole source of finding 1.

jkitchin:
All three fixed in 216a755. Each verified against the code and reproduced before fixing — thanks, finding 1 in particular was a real regression I'd have shipped.

### 1. The `unknown_lints` trap — confirmed, and worse than the file it's in

Reproduced exactly, by installing 1.96.0:

```
$ cargo +1.96.0 clippy --lib -- -D warnings
error: unknown lint: `clippy::chunks_exact_to_as_chunks`
  --> src/dense/schur_kernel.rs:41:10
```

You're right that this blocks **every** local commit for anyone below 1.98, not just commits touching that file — and that CI tracking `stable` guarantees it stays invisible. My "clippy clean" claim in the PR body was only ever true on the toolchain this session had *just* updated to in order to reproduce the original failure. That's the second time in this PR that a stale-vs-current toolchain gap produced a false green, in the opposite direction.

Fixed with `#[allow(unknown_lints)]` above it, as you suggested. Now clean on both:

```
cargo +1.96.0 clippy --all-targets -- -D warnings       -> clean
cargo clippy --workspace --all-targets -- -D warnings   -> clean
```

### 2. `with_scaling` mid-escalation — confirmed

Real, and it is the lazy-snapshot failure mode relocated to the window R6 doesn't cover. `with_scaling` now refreshes `quality_baseline.scaling` when a snapshot is live, on the reading that the call re-states what the caller's baseline *is*.

I took refresh over clear: clearing drops the pivot-threshold half of the snapshot too, so a later `reset_quality` would leave the pivot escalated and report nothing undone — trading a wrong scaling for a silently wrong pivot.

### 3. Phantom undo — confirmed

`reset_quality` now compares the snapshot against current values and returns whether anything actually differs; the level is re-baselined either way. Doc comment and CHANGELOG restate the contract precisely, including the `0.5^0.75` clamp case.

### One thing worth flagging: findings 2 and 3 compose, and my first test was wrong about it

R7 as I first wrote it asserted `reset_quality() == true` after pinning a strategy post-escalation. Once finding 3 is fixed that's **false, and correctly so** — `with_scaling` had already superseded the only rung that fired (stage 1 touches scaling only), so nothing was left to restore. The test was wrong, not the code. Split rather than papered over:

- **R7** — escalates twice (scaling *and* pivot) so the return value stays meaningful; pins that the pinned strategy survives the reset.
- **R7b** — pins the composed behaviour head-on: pin supersedes the only rung → `false`, strategy stays pinned, level still re-baselined.
- **R8** — a rung that moved nothing is not reported as undone.

Both findings reproduced RED first (`assertion failed: s.reset_quality()`, `no parameter moved, so nothing was undone`).

```
cargo test --workspace -> 1199 passed, 0 failed
cargo fmt --all --check -> clean
pytest python/tests/test_basic.py -> 16 passed (maturin develop)
```

### On the scope note

Agreed, and it's the right call to have flagged it — 9c5dfac is unrelated to #192 and is the sole source of finding 1. I ported it because the toolchain bump turns `check` red on any PR opened against `main` today, so this PR couldn't go green without it. Happy to split it into its own PR if you'd rather review it separately; it's a clean `git cherry-pick` of 9c5dfac + the finding-1 hunk of 216a755, and this branch would rebase to just the feature work.


---

jkitchin:
Catching up on the downstream comment — it didn't reach me as an event, only the later review did, so I found it by polling the comment list. Flagging that so you know the wake stream isn't a complete record.

All three contract asks are done in 81a1785. Documentation and CHANGELOG only; no functional change.

### The three properties are now contract

Each was already true and already pinned, but asserted only in the PR description — which is exactly the "implementation detail, not contract" gap you're pointing at. They're now in a `# Guarantees` section on the `reset_quality` doc comment and in the CHANGELOG, each naming the test that pins it and saying a change that breaks one is a breaking change:

1. **Never calling `reset_quality` → byte-identical behaviour to 0.17.0.** Pinned by U1–U5, which predate the method and pass unchanged.
2. **`reset` → `increase` retraces a fresh `Solver`'s rungs.** Pinned by R4 (records the traversal *before* any reset exists in the state, asserts the second equals it) and R6.
3. **Re-baselining costs no symbolic re-analysis.** Written up explicitly as a *cost* guarantee rather than a correctness one, with your once-per-IPM-iteration call site as the reason it matters. Pinned by I9 via `symbolic_call_count`.

The baseline-binary-not-a-spec rationale is the convincing part — a property that's only true by accident is one a refactor moves silently, and your sweep would surface it as unexplained drift long after the cause. Worth having in the source.

### Correction to something I'd been repeating

Your point about the firing cap is one I'd gotten wrong and had been restating in the PR body: I'd been treating #192's measurements as evidence for what the reset *does*. They aren't. The cap left the escalation in force for the remainder of the solve, which is precisely why allowing only the first firing still lost the leg — so a reset at the matrix boundary expresses something the cap could not measure, in either direction.

I've been careful in the docs not to claim the reset recovers `square_flowsheet_resto`; the claim is only that it makes the lifetime expressible. Agreed that the corpus re-run is the thing that settles it.

### On the guard

Noted, and that retires it from my side — I'd deferred it pending evidence, and "we'd rather own our own retry loop" is the answer. The note in `dev/tried-and-rejected.md` stands as a record of the reasoning rather than an open question.

### Sequencing and the corpus run

Both your calls, and I've not acted on either:

- **#191 + this together for 0.18.0** — makes sense given both touch `src/numeric/solver.rs` and it saves you a fixture sweep. This branch is green and mergeable whenever you want it; say the word if you'd like me to rebase it on #191 (or vice versa) rather than leaving the ordering to the merge.
- **Corpus run before the release rather than the merge** — agreed, and that matches what the session checkpoint already lists as the next session's first item. This container has no corpus, so I can't close it here.

Still open from my previous comment: whether you'd like 9c5dfac split into its own PR. No action needed if you're happy with it riding along.

```
cargo test --workspace -> 1199 passed, 0 failed
cargo clippy --all-targets -- -D warnings -> clean on 1.96 and 1.98
cargo fmt --all --check -> clean
cargo doc --no-deps -p feral -> 40 warnings, same as origin/main (all pre-existing)
```


---

Issue #65, opened 2026-06-03
Auto scaling reports wrong inertia (spurious zero pivots) on ill-conditioned KKT; Mc64Symmetric fixes it (false-infeasible in pounce)


jkitchin:
## Summary

On an ill-conditioned symmetric-indefinite KKT system, FERAL 0.9.0's **default `Auto` scaling** mis-factors the matrix and reports a **wrong inertia with ~100 spurious zero pivots**, where MA27/MA57 (and a dense `eigvalsh` reference) report the correct full-rank inertia. Switching to **`ScalingStrategy::Mc64Symmetric` fixes it exactly.** Because the wrong inertia is reported as `Singular`/rank-deficient, the consuming IPM ([pounce](https://github.com/jkitchin/pounce)) takes a bad regularized step at iteration 0 and falsely declares the problem infeasible.

This is distinct from #63 (that one is *correct* inertia but excessive backward error flooring feasibility; ordering-dependent). This one is **wrong inertia**, **scaling-dependent**, and surfaces as a hard `Singular` misclassification.

Observed with `feral 0.9.0` via the `pounce-feral` backend. Downstream symptom: pounce reports `discs` and `sawpath` (Vanderbei/CUTEst) as "Converged to a point of local infeasibility" while Ipopt-MA57 finds the optimum (pounce#95 Lead 2b).

## The matrix

The augmented KKT at pounce's **iteration 0** on `sawpath`:

- `n = 1575` (789 primal incl. slacks + 786 dual), `nnz = 5684` lower-triangle triplets
- **genuinely ill-conditioned**: `cond ≈ 3.7e20`, with 3 eigenvalues at the ~`1e-15` noise floor and a dense cluster at ~`1e-8`. So it is *near*-singular but effectively full rank (numpy `matrix_rank` = 1572 at default tol; 0 eigenvalues below `1e-20`).
- **ground-truth inertia (dense `numpy.linalg.eigvalsh`)**: `(789, 786, 0)`.
- **Ipopt MA27/MA57**: factors with `delta = 0` (zero perturbation), 1 trial factorization, inertia `(789, 786, 0)`, and goes on to solve the NLP in ~11 iterations.

## FERAL behavior (the bug)

Factoring the **same** matrix with `feral::Solver`, varying only the scaling strategy:

| `ScalingStrategy` | reported inertia (pos, neg, zero) | `min_pivot_magnitude` | verdict |
|---|---|---|---|
[...]

PR #195, opened 2026-08-30
feat: cooperative cancellation for Solver::factor (#194)

### LINKED ISSUE #194: factor() cannot be cancelled: a single factorization larger than the caller's whole budget makes a wall-clock limit unenforceable
## Summary

`Solver::factor` cannot be cancelled once it is running, and there is no
surface through which a caller could ask. For a host that enforces a
wall-clock budget, that makes the budget unenforceable whenever a single
factorization is larger than the whole budget — the host cannot check a
deadline it never gets control back to check.

Measured in pounce: on `emfl050_5_5` (5675 variables, 5625 constraints), a
`max_wall_time=5` solve returned `TIME_LIMIT` after **48.8 seconds**. The
between-operation check passed while still under budget, one factorization
then ran ~44 s uninterrupted, and only the *next* check tripped. feral's own
#8 records the mechanism: a delayed-pivot cascade turns "an otherwise
sub-second problem" into an 87 s factorization.

## Why pounce cannot fix this in its own tree

Verified against the 0.17.0 source (`src/numeric/solver.rs`,
`src/numeric/factorize.rs`):

- `pub fn factor(&mut self, matrix: &CscMatrix, check_inertia: Option<Inertia>) -> FactorStatus`
  takes no interrupt, deadline, cancel flag, or progress callback.
- `FactorStatus` is `Success | Singular | WrongInertia | FatalError`, with no
  variant that could carry an early return.
- `NumericParams` and the `with_*` builders expose nothing equivalent.
- `DelayBudgetExceeded` exists but is a symbolic-time **column-count** budget
  on the delayed-pivot cascade, not a wall-clock one, and it is already armed
  by default. It did not save the case above.

A thread-based watchdog on the pounce side was considered and rejected: our
backend handle holds non-`Send`, stateful symbolic-cache buffers reused
across calls, so abandoning a factorization mid-flight either leaves those
buffers being mutated by an orphaned thread (unsound) or forces a fresh
`Solver` per factorization, destroying the symbolic-cache reuse the IPM
depends on.

Pounce has already built the half it *can* build. `PdFullSpaceSolver`
measures each factorization's cost and refuses to *start* one it predicts
cannot finish inside the remaining budget, which bounds the overshoot
proactively in the "several factorizations, one budget" regime. It provably
cannot close this case: the offending factorization is the **first** one,
before any estimate exists. There is nothing to predict from.

## What would fix it

Cooperative cancellation polled at a coarse-but-frequent granularity — per
supernode, or every N eliminated columns. Often enough to bound the overshoot
well under a second, rare enough to stay off the hot inner kernels. Polling a
shared `AtomicBool` rather than a clock keeps feral clock-agnostic and leaves
the wall-vs-CPU-vs-budget policy with the caller, which is where it belongs.

Matching the existing builder convention on `Solver`:

```rust
impl feral::Solver {
    /// Arm cooperative cancellation for subsequent `factor` calls.
    /// `None` (default) keeps the uninterruptible path at zero overhead.
    pub fn with_interrupt(self, flag: Arc<AtomicBool>) -> Self;
}

pub enum FactorStatus {
    // ...
    /// The interrupt flag was observed set. Factors are left invalid.
    Interrupted,
}
```

An optional `NumericParams` field would work equally well; the builder is
suggested only because `with_parallel` / `with_scaling` / `with_cascade_break`
already establish that shape.

**Contract we would rely on:** on `Interrupted` the `Solver`'s factors are
left invalid, exactly as after any failed factor; the caller must not `solve`
against them; a subsequent `factor` re-runs cleanly. No partial results
promised, and no guarantee about *when* within a supernode the check fires.

## Wiring on the pounce side once it ships

Recorded here so the API shape can be judged against a real consumer:

1. `pounce_linsol::SparseSymLinearSolverInterface` grows an optional
   `set_interrupt(&mut self, Option<Arc<AtomicBool>>)` (default no-op), and
   `ESymSolverStatus` grows an `Interrupted` variant that
   `TSymLinearSolver::multi_solve` forwards.
2. `pounce_feral::FeralSolverInterface` arms the `Solver` and maps
   `FactorStatus::Interrupted -> ESymSolverStatus::Interrupted`.
3. `PdFullSpaceSolver` owns the `Arc<AtomicBool>` and sets it from a
   lightweight watchdog when its `Deadline` first trips.
   `ESymSolverStatus::Interrupted` then returns `false` from `solve_once` and
   routes through the existing post-KKT deadline check exactly like the
   between-operation abort does today — **no new termination path**.

With that, a `max_wall_time=5` solve returns in ~5 s plus one supernode
instead of 48.8 s.

## Not asking for

A deadline or clock inside feral, progress reporting, partial factors, or any
guarantee of promptness beyond "checks at supernode boundaries". A flag and a
status variant are enough.

Full pounce-side note, including the rejected alternatives:
`dev-notes/feral-factor-interrupt.md`. Filed from the pounce side (pounce
gh#254); happy to supply `emfl050_5_5` or the KKT systems if useful.

jkitchin:
Closes #194.

## What

`Solver::factor` can now be cancelled by a caller-owned `Arc<AtomicBool>`.

- `Solver::with_interrupt(flag)` — the builder the issue proposed.
- `Solver::set_interrupt(Option<flag>)` — the stated consumer holds its `Solver` behind `&mut self` (`SparseSymLinearSolverInterface::set_interrupt`), where a consuming builder cannot be called; it is also the only way to disarm.
- `Solver::interrupt()` — read back what is armed.
- `FactorStatus::Interrupted` and `FeralError::Interrupted`.

The flag itself is `BunchKaufmanParams::interrupt`, so direct users of `factorize_multifrontal*` and of the dense frontal factor get the same mechanism.

## Why

`factor` could not be cancelled once running and offered no surface through which a caller could ask, so a host enforcing a wall-clock budget could not enforce it whenever a single factorization was larger than the whole budget — it never regained control to check its own deadline. Measured in pounce on `emfl050_5_5` (5675 variables, 5625 constraints): a `max_wall_time=5` solve returned `TIME_LIMIT` after **48.8 s**, because the between-operation check passed while under budget and one factorization then ran ~44 s uninterrupted.

Every claim in the issue was checked against this tree before designing, rather than taken at face value: `factor`'s signature, `FactorStatus`' variants, `NumericParams`' fields, `DelayBudgetExceeded`'s actual semantics (a symbolic-time column-count budget, not wall-clock — as reported), and the cross-call mutable state (`workspace`, `last_symbolic`, `permute_cache`, `parallel_pool`) that makes a caller-side watchdog unsound. `dev/tried-and-rejected.md` has no prior attempt at this feature.

## One thing the issue asked for would not have fixed the issue

Supernode-boundary polling alone leaves the reported case intact. feral's own #8 measured a delayed-pivot cascade concentrating 118k delayed pivots into three ~14k-column expanded root fronts, so an 87 s factorization is dominated by a handful of *individual fronts*; a check that fires only between supernodes returns "budget plus one 44 s supernode", which is the reported bug again.

So there are two poll granularities: supernode boundaries in both multifrontal drivers, **and** the `while k < ncol_eff` panel loop in the dense frontal factor (blocked and unblocked paths). That loop is still one level out from the hot kernels — every polled iteration guards at least `O(nrow)` of work. The published contract stays the weaker of the two: supernode boundaries, and within a supernode dense panel boundaries, with no promise about when within a panel.

## Design notes

- **A flag, not a deadline.** feral holds no clock; wall-vs-CPU-vs-budget policy stays with the caller. `Ordering::Relaxed` is sufficient — the flag carries no data dependency, and a missed observation costs one more panel.
- **feral only ever reads the flag.** Re-arming after an interrupt is the caller's `store(false)`.
- **`Interrupted` is a `FeralError` variant**, so both drivers' existing unwind carries the cancellation with no new control flow: the sequential `?` out of the supernode loop, and the parallel `first_error` plus the fast-exit at the top of `run_parallel_task`, which drains the scope without starting further work.
- **The flag lives on `BunchKaufmanParams`, not `NumericParams`.** The drivers hold `params.bk` and the dense frontal factor is handed *only* `&BunchKaufmanParams`, so one field is readable from every poll site with no sync step — exactly the shape that left `NumericParams::fma` a silent no-op until finding N1.

Full rationale in `dev/research/issue-194-cooperative-interrupt.md` and the 2026-08-30 entry in `dev/decisions.md`.

## Contract

On `Interrupted` the factors are left invalid exactly as after any failed factor — `inertia()` is `None`, `solve()` returns `NoFactor` — and a subsequent `factor` re-runs cleanly once the caller clears the flag. No partial results are promised. Only the numeric factorization is covered: the symbolic analysis on the first factor of a new pattern is not interruptible (for an IPM refactoring a fixed pattern that is a cache hit, which is the case the flag exists for).

## Evidence

`tests/issue194_factor_interrupt.rs`, 8 tests, all passing. In `--release`, warm solver (symbolic cached, the IPM case), 5-point Laplacian `n = 122500`:

| driver | un-interrupted | flag set at | returned `Interrupted` |
|---|---|---|---|
| sequential | 263.2 ms | 26.3 ms | 34.8 ms |
| parallel | 153.5 ms | 15.3 ms | 25.5 ms |

Overshoot past the request is ~9–10 ms, most of it the un-polled prologue (scaling and permute), against a factorization that would otherwise have run to completion.

The tests also pin the rest of the contract: an unarmed `Solver` is unaffected; an armed-but-clear flag still succeeds with the baseline inertia (so the flag is polled, not merely present); an interrupted factor stores no factor and `solve` returns `NoFactor`; clearing the flag lets the next `factor` re-run cleanly and reproduce the un-interrupted inertia — including after a *mid-flight* abort, which is the case a naive implementation (poisoned workspace, stale contribution blocks) would break; and feral never writes the caller's flag.

Full suite: 100 test binaries, 0 failures. `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean for `feral` and for `feral-diagnostics`; `cargo check` clean for the Python bindings.

## Two things to flag for review

**1. The benchmark gate is unmet, and this PR does not claim otherwise.** The corpus is not present in the container this was developed in, so `cargo run --bin bench --release` ran only the 8 synthetic matrices and both exit-partition tables read `N/A`. Session 2026-08-19-05's p90s (dense 1.58 / 2.00, sparse 1.58 / 1.58) stand unconfirmed. In its place, the poll overhead was measured directly with a paired A/B against `origin/main` (same probe source, best-of-10 warm refactors, arms alternated):

```
                     branch                main (9b9e882)
grid_laplacian_350   203.8 / 197.8 ms      222.8 / 203.5 ms   sequential
grid_laplacian_350    94.2 /  92.3 ms       94.4 /  92.2 ms   parallel
tridiag_200k          57.7 /  57.3 ms       59.4 /  59.2 ms   sequential
tridiag_200k          56.3 /  56.9 ms       56.8 /  54.6 ms   parallel
```

No measurable overhead — every poll site short-circuits on an `Option::is_some` branch and touches no atomic when unarmed. But the honest claim is "no regression detectable on this host", not "zero cost proven": the *first*, un-alternated A/B block read `main` at 283–291 ms on `grid_laplacian_350` sequential, an apparent 40% regression that turned out to be container drift and disappeared on re-running interleaved. Someone should re-run `bench` with the corpus mounted.

**2. C and Python can name the status but cannot arm the flag.** `FERAL_INTERRUPTED = 4` (`src/capi.rs`, `feral-ipopt-shim/include/feral_capi.h`) and `FactorStatus.INTERRUPTED` are defined so the exhaustive matches stay honest, but neither surface has an arming API — #194's consumer is Rust. Whether to add one (`feral_set_interrupt`, or a small `feral.InterruptFlag`; `factor` already runs under `py.allow_threads`) is recorded as a follow-up in the session checkpoint.

## Blast radius

Adding a variant to each of `FactorStatus` and `FeralError` broke exhaustive matches in 4 in-tree integration tests, 10 `feral-diagnostics` probe binaries (built by CI with `-D warnings`, and outside the root workspace default set, so the root `cargo clippy --all-targets` does not see them), and the Python bindings. All were given **explicit arms rather than wildcards**, deliberately: the next variant should break the same builds rather than be silently absorbed. An unarmed `Solver` is unaffected in behaviour and in performance.

https://claude.ai/code/session_01Jy5x8pZLRm8GDaSNXMokLP

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jy5x8pZLRm8GDaSNXMokLP)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,12 +45,35 @@ All notable changes to FERAL will be documented in this file.
   returned at 25.5 ms. A paired A/B against the previous release found no
   measurable cost when unarmed — every poll site short-circuits on an
   `Option::is_some` branch and touches no atomic.
-- **Additive.** An unarmed `Solver` is unaffected in behaviour and in
-  performance. `FactorStatus` and `FeralError` each gain one variant, so
-  exhaustive `match`es over them need a new arm. The C ABI defines
-  `FERAL_INTERRUPTED = 4` and the Python bindings define
-  `FactorStatus.INTERRUPTED`; neither surface can arm the flag yet, so on both
-  the code is reserved rather than reachable.
+- **BREAKING for exhaustive `match`es; additive at runtime.** An unarmed
+  `Solver` is unaffected in behaviour and in performance — nothing that
+  already worked changes what it does. But `FactorStatus` and `FeralError`
+  each gain one variant and neither is `#[non_exhaustive]`, so every
+  downstream `match` over them that lists all arms stops compiling until a
+  new arm is added. In-tree this broke 4 integration tests, 10
+  `feral-diagnostics` probe binaries, and the Python bindings; a downstream
+  crate will see the same. Each was given an explicit arm rather than a
+  wildcard, deliberately: the next variant should break the same builds
+  rather than be silently absorbed. The C ABI defines `FERAL_INTERRUPTED = 4`
+  and the Python bindings define `FactorStatus.INTERRUPTED`; neither surface
+  can arm the flag yet, so on both the code is reserved rather than
+  reachable.
+- **One defect found in review and fixed before merge.** The issue-#65 MC64
+  rescue re-factors with `Mc64Symmetric` when the first factorization reports
+  the singular signature, and its non-adoption arm was a bare `_` that also
+  caught `Err(_)`. An interrupt observed during that *retry* was therefore
+  swallowed: `factor` returned `Success` rather than `Interrupted`, and
+  `mc64_retry_not_adopted` latched. Because that latch is keyed on the
+  pattern and cleared only on a pattern change, a single cancellation would
+  have suppressed the MC64 rescue for every later factor of the same pattern
+  — for an interior-point host, the rest of the solve — reporting unrescued
+  inertia where it would otherwise have recovered. The window was not narrow:
+  it was the entire duration of a second full factorization. Fixed by
+  propagating `Err(FeralError::Interrupted)` explicitly and leaving the latch
+  disarmed, since a cancellation is not evidence about MC64 — the retry never
+  finished. Regression test: `tests/issue194_interrupt_during_mc64_retry.rs`,
+  verified to fail before the fix (observed `Success` on a cancelled call).
+
 ### Added - iterative refinement can be told what "converged" means (issue #190)
 
 - **The problem.** The refinement loop's only convergence test was a hardwired
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,56 @@
 
 All notable changes to FERAL will be documented in this file.
 
+## [Unreleased]
+
+### Added — `Solver::factor` can now be cancelled (issue #194)
+
+- **What changed.** A caller-owned `Arc<AtomicBool>` can be armed on a
+  `Solver` with `with_interrupt(flag)` (builder) or `set_interrupt(Some(flag))`
+  (through a `&mut` borrow; `None` disarms), and read back with `interrupt()`.
+  While armed, `factor` polls the flag and, on the first observation of `true`,
+  stops and returns the new `FactorStatus::Interrupted`. The underlying
+  free-function drivers surface it as the new `FeralError::Interrupted`, and
+  the flag itself is `BunchKaufmanParams::interrupt`, so direct users of
+  `factorize_multifrontal*` and of the dense frontal factor get the same
+  mechanism.
+- **Why.** `factor` could not be cancelled once running and offered no surface
+  through which to ask, so a host enforcing a wall-clock budget could not
+  enforce it whenever a single factorization was larger than the whole budget —
+  it never regained control to check its own deadline. Measured in pounce on
+  `emfl050_5_5` (5675 variables, 5625 constraints): a `max_wall_time=5` solve
+  returned `TIME_LIMIT` after **48.8 s**, because the between-operation check
+  passed while under budget and one factorization then ran ~44 s uninterrupted.
+  A watchdog on the caller's side cannot fix this: `Solver` holds workspace,
+  symbolic and permute caches across calls, so abandoning a factorization
+  mid-flight either orphans mutable state or forces a fresh `Solver` per call
+  and destroys the symbolic reuse an interior-point host depends on.
+- **Contract.** On `Interrupted` the factors are left invalid exactly as after
+  any failed factor — `inertia()` is `None`, `solve()` returns `NoFactor` — and
+  a subsequent `factor` re-runs cleanly once the caller clears the flag. No
+  partial results are promised. Polling happens at supernode boundaries and, in
+  a supernode, at dense panel boundaries; there is no guarantee about *when*
+  within a panel the check fires. Only the numeric factorization is covered:
+  the symbolic analysis on the first factor of a new pattern is not
+  interruptible.
+- **Caller-owned, and no clock.** feral only ever *reads* the flag, so
+  re-arming after an interrupt is the caller's `store(false)`. Taking a flag
+  rather than a deadline keeps feral clock-agnostic and leaves
+  wall-vs-CPU-vs-budget policy where it belongs.
+- **Measured.** In `--release` on a warm solver (symbolic cached, the
+  interior-point case), 5-point Laplacian `n = 122500`: sequential driver,
+  263.2 ms un-interrupted, flag set at 26.3 ms, `Interrupted` returned at
+  34.8 ms; parallel driver, 153.5 ms un-interrupted, flag set at 15.3 ms,
+  returned at 25.5 ms. A paired A/B against the previous release found no
+  measurable cost when unarmed — every poll site short-circuits on an
+  `Option::is_some` branch and touches no atomic.
+- **Additive.** An unarmed `Solver` is unaffected in behaviour and in
+  performance. `FactorStatus` and `FeralError` each gain one variant, so
+  exhaustive `match`es over them need a new arm. The C ABI defines
+  `FERAL_INTERRUPTED = 4` and the Python bindings define
+  `FactorStatus.INTERRUPTED`; neither surface can arm the flag yet, so on both
+  the code is reserved rather than reachable.
+
 ## [0.17.0] - 2026-08-19
 
 ### Fixed — the tree-parallel solve no longer runs on trees too thin to pay for it (issue #175)
--- a/crates/feral-diagnostics/src/bin/diag_issue50_auto_validate.rs
+++ b/crates/feral-diagnostics/src/bin/diag_issue50_auto_validate.rs
@@ -127,6 +127,10 @@ fn run_auto(matrix: &CscMatrix) -> (Option<usize>, Option<u64>, String, String) 
         FactorStatus::Singular => "singular",
         FactorStatus::WrongInertia { .. } => "wrong_inertia",
         FactorStatus::FatalError(_) => "fatal",
+        // Issue #194: unreachable here (this probe never arms an
+        // interrupt flag), but labelled rather than wildcarded so a
+        // future variant cannot be silently folded into another.
+        FactorStatus::Interrupted => "interrupted",
     };
     let (resolved, nnz_l) = match solver.factors() {
         Some(f) => (format!("{:?}", f.resolved_method), Some(f.factor_nnz())),
--- a/crates/feral-diagnostics/src/bin/diag_issue50_numeric_inventory.rs
+++ b/crates/feral-diagnostics/src/bin/diag_issue50_numeric_inventory.rs
@@ -207,6 +207,7 @@ fn run_one(matrix: &CscMatrix, method: OrderingMethod) -> (Option<usize>, Option
         FactorStatus::Singular => "singular",
         FactorStatus::WrongInertia { .. } => "wrong_inertia",
         FactorStatus::FatalError(_) => "fatal",
+        FactorStatus::Interrupted => "interrupted",
     };
     let nnz_l = solver.factors().map(|f| f.factor_nnz());
     let factor_us = if matches!(status, FactorStatus::Success) {
--- a/crates/feral-diagnostics/src/bin/diag_near_singular_sweep.rs
+++ b/crates/feral-diagnostics/src/bin/diag_near_singular_sweep.rs
@@ -112,6 +112,7 @@ fn main() {
             FactorStatus::Singular => "Singular",
             FactorStatus::WrongInertia { .. } => "WrongInertia",
             FactorStatus::FatalError(_) => "FatalError",
+            FactorStatus::Interrupted => "Interrupted",
         };
 
         let inertia = solver.inertia().cloned().unwrap_or(feral::Inertia {
@@ -201,6 +202,7 @@ fn main() {
             FactorStatus::Singular => "Singular",
             FactorStatus::WrongInertia { .. } => "WrongInertia",
             FactorStatus::FatalError(_) => "FatalError",
+            FactorStatus::Interrupted => "Interrupted",
         };
         let inertia = solver.inertia().cloned().unwrap_or(feral::Inertia {
             positive: 0,
--- a/crates/feral-diagnostics/src/bin/diag_small_sparse_inventory.rs
+++ b/crates/feral-diagnostics/src/bin/diag_small_sparse_inventory.rs
@@ -160,6 +160,7 @@ fn run_one(matrix: &CscMatrix, method: OrderingMethod) -> (Option<usize>, Option
         FactorStatus::Singular => "singular",
         FactorStatus::WrongInertia { .. } => "wrong_inertia",
         FactorStatus::FatalError(_) => "fatal",
+        FactorStatus::Interrupted => "interrupted",
     };
     let nnz_l = solver.factors().map(|f| f.factor_nnz());
     let factor_us = if matches!(status, FactorStatus::Success) {
--- a/crates/feral-diagnostics/src/bin/phase0_cb_on_revalidation.rs
+++ b/crates/feral-diagnostics/src/bin/phase0_cb_on_revalidation.rs
@@ -130,6 +130,10 @@ fn run_case(case: &Case) -> Outcome {
             pass: false,
             detail: format!("FAIL FatalError({e:?})  factor={dt:.3}s"),
         },
+        FactorStatus::Interrupted => Outcome {
+            pass: false,
+            detail: format!("FAIL Interrupted  factor={dt:.3}s"),
+        },
     }
 }
 
--- a/crates/feral-diagnostics/src/bin/probe_issue63_nearsingular.rs
+++ b/crates/feral-diagnostics/src/bin/probe_issue63_nearsingular.rs
@@ -116,6 +116,7 @@ fn status_short(s: &FactorStatus) -> &'static str {
         FactorStatus::Singular => "Singular",
         FactorStatus::WrongInertia { .. } => "WrongInertia",
         FactorStatus::FatalError(_) => "FatalError",
+        FactorStatus::Interrupted => "Interrupted",
     }
 }
 
--- a/crates/feral-diagnostics/src/bin/probe_issue65_scaling.rs
+++ b/crates/feral-diagnostics/src/bin/probe_issue65_scaling.rs
@@ -46,6 +46,7 @@ fn status_short(s: &FactorStatus) -> String {
             format!("WrongInertia(act {actual:?} exp {expected:?})")
         }
         FactorStatus::FatalError(e) => format!("FatalError({e:?})"),
+        FactorStatus::Interrupted => "Interrupted".to_string(),
     }
 }
 
--- a/crates/feral-diagnostics/src/bin/probe_kkt_replay.rs
+++ b/crates/feral-diagnostics/src/bin/probe_kkt_replay.rs
@@ -143,6 +143,7 @@ fn main() {
             ),
             FactorStatus::Singular => "Singular".to_string(),
             FactorStatus::FatalError(e) => format!("Fatal: {e:?}"),
+            FactorStatus::Interrupted => "Interrupted".to_string(),
         };
         println!("{i:>4}  {elapsed:>10.3}  {p:>10}  {n:>10}  {z:>10}  {label}");
     }
--- a/crates/feral-diagnostics/src/bin/probe_pinene_issue38_fix.rs
+++ b/crates/feral-diagnostics/src/bin/probe_pinene_issue38_fix.rs
@@ -75,6 +75,7 @@ fn main() {
             ),
             FactorStatus::Singular => "Singular".to_string(),
             FactorStatus::FatalError(e) => format!("Fatal: {e:?}"),
+            FactorStatus::Interrupted => "Interrupted".to_string(),
         };
 
         println!("{i:>4}  {elapsed:>10.3}  {p:>10}  {n:>10}  {z:>10}  {label}");
--- a/crates/feral-diagnostics/src/bin/probe_warm_cascade.rs
+++ b/crates/feral-diagnostics/src/bin/probe_warm_cascade.rs
@@ -80,6 +80,7 @@ fn factor_one(
         FactorStatus::WrongInertia { .. } => "WrongInertia",
         FactorStatus::Singular => "Singular",
         FactorStatus::FatalError(_) => "Fatal",
+        FactorStatus::Interrupted => "Interrupted",
     };
     println!("  {label:<8} {secs:>10.3} s  {tag}");
     secs
--- a/dev/context.md
+++ b/dev/context.md
@@ -1,70 +1,335 @@
 # FERAL Context (auto-generated)
 
-Generated: 2026-08-20T02:05:49Z
+Generated: 2026-08-30T00:31:02Z
 
 ## Latest Session
-File: dev/sessions/2026-08-19-04.md
+File: dev/sessions/2026-08-19-05.md
 ```
-# Session 2026-08-19-04
-
-## BENCHMARK NUMBERS ARE NOT COMPARABLE TO LAST SESSION
-
-Reported first, per the hard rule in CLAUDE.md — and for the same reason
-as 2026-08-19-02: `cargo run --bin bench --release` in this container
-finds only the 8 synthetic matrices. The external corpus and the
-MUMPS/SPRAL oracle timings are absent, so both Phase 2.8.1 exit
-partitions report `N/A` and **no comparison against 2026-08-15-02's
-1.61 / 2.00 / 1.67 / 1.67 is possible from this run**. I am not claiming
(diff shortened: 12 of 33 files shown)


jkitchin:
The `check` job is red, and it is not this PR's failure — `main` fails identically. Standing down on it rather than widening this PR, with the evidence and a proposed fix below.

## What is failing

`pre-commit run --all-files` → the `cargo-clippy` hook:

```
error: unneeded late initialization
    --> crates/feral-ordering-core/src/quotient_graph/algo.rs:1207:13
    = note: `-D clippy::needless-late-init` implied by `-D warnings`
error: could not compile `feral-ordering-core` (lib) due to 1 previous error
```

## Why it is not this PR's

- This PR does not touch that crate at all: `git diff --stat origin/main...HEAD -- crates/feral-ordering-core` is empty.
- It reproduces on `main` untouched. Checked out `origin/main` at 9b9e882 in a clean worktree and ran the CI toolchain against it:

  ```
  $ cargo +1.98.0 clippy -- -D warnings
  error: unneeded late initialization
      --> crates/feral-ordering-core/src/quotient_graph/algo.rs:1207:13
  error: could not compile `feral-ordering-core` (lib) due to 1 previous error
  ```

This is toolchain drift, not a regression. `main`'s last CI run (run 32364055588, 2026-08-20) was green; rustc/clippy 1.98.0 was released 2026-08-18 and the hosted runner has since picked it up. `needless_late_init` fires on pre-existing code that no longer passes the newer lint set. Nothing in the tree pins a toolchain, so every PR opened from now on gets this failure until it is fixed on `main`.

Re-running the job would not help — this is a deterministic lint, not a flake — so I have not spent a re-run on it.

## No fix exists yet to port, so here is the proposed patch

Two separate lint families are involved. The first is the hard error above, and it is a one-hunk mechanical change (`crates/feral-ordering-core/src/quotient_graph/algo.rs`, ~line 1207): hoist the declaration into the `if`/`else` expression and drop the assignments from the branches.

```diff
-            let rmf: f64;
             let deg_i = ws.degree[i];
-            if (deg_i as usize) + (degme_i as usize) > nleft {
+            let rmf: f64 = if (deg_i as usize) + (degme_i as usize) > nleft {
                 // Saturated branch. RMF1 uses original DEG.
                 let deg_f = deg_i as f64;
                 let rmf1 = deg_f * (deg_f - 1.0 + 2.0 * degme_i as f64) - ws.wf[i] as f64;
                 let new_deg = (nleft as i32) - nvi_i;
                 ws.degree[i] = new_deg;
                 let nd = new_deg as f64;
                 let rmf_new =
                     nd * (nd - 1.0) - (degme_i - nvi_i) as f64 * (degme_i - nvi_i - 1) as f64;
-                rmf = rmf_new.min(rmf1);
+                rmf_new.min(rmf1)
             } else {
                 let deg_f = deg_i as f64;
                 ws.degree[i] = deg_i + degme_i - nvi_i;
-                rmf = deg_f * (deg_f - 1.0 + 2.0 * degme_i as f64) - ws.wf[i] as f64;
+                deg_f * (deg_f - 1.0 + 2.0 * degme_i as f64) - ws.wf[i] as f64
             };
```

**That is not sufficient on its own.** Clippy aborts at the first crate that fails, so it was hiding a second family. With the patch above applied to `main`, `cargo +1.98.0 clippy --all-targets -- -D warnings` surfaces 10 more sites, all in one file:

```
6 × src/dense/schur_kernel.rs: using `chunks_exact` with a constant chunk size
4 × src/dense/schur_kernel.rs: using `chunks_exact_mut` with a constant chunk size
   (clippy::chunks_exact_to_as_chunks — lines 391, 392, 467, 468, 469, 603, 604, …)
```

Those are the only two families on `main` at 1.98.0; there is nothing else hiding behind them.

## Suggested handling

I would rather not carry either fix in this PR. The `chunks_exact` sites are in the hot packed-SIMD trailing-update kernel, which carries the cross-arch bit-exactness contract that `golden_bits` and `fma_vs_nofma_panel_kernels_within_n_elim_ulps` guard. `chunks_exact(4)` → `as_chunks::<4>()` should be semantically identical, but it is a change to numerical kernel code that deserves its own PR and its own golden-digest run, not a ride-along in a cancellation feature.

So the suggestion is a small toolchain-drift PR against `main` carrying both families, after which this one goes green on a rebase. Worth considering in the same PR: pinning a toolchain (`rust-toolchain.toml`) or dropping `-D warnings` for style lints in CI, so the next stable bump does not turn every open PR red for reasons unrelated to any of them.

Everything else on this PR is green: `stress-smoke`, all three `Test wheel` jobs, `llvm-cov → Codecov`, and both Codecov checks.

---

jkitchin:
## Review

The design is right and the write-up is unusually honest — flagging the unmet
benchmark gate, and flagging that the first un-alternated A/B block was
container drift rather than reporting the 40% "regression", is exactly the
standard this repo asks for. Two poll granularities rather than one is the
correct call, and the issue-#8 justification for it is concrete.

I checked the mechanism itself and it holds up:

- Ordering escalation (`src/numeric/solver.rs:1558`) sits on the `Ok` path, so
  `Interrupted` cannot drive the recursive re-`factor`.
- `run_parallel_task`'s `task_res = Err(..); break` lands in the pre-existing
  `Err` arm (`src/numeric/factorize.rs:3919`), which sets `first_error` and
  skips the parent enqueue — and it is a `break`, not a `return`, so the
  `local_contribs` restore into the workspace still runs. That distinction
  matters and it was gotten right.
- Explicit arms instead of wildcards across the blast radius is the right
  trade, for the stated reason.

One defect, and it is worth fixing before merge.

### The MC64 rescue swallows `Interrupted`, and latches

`src/numeric/solver.rs:1359`:

```rust
match retry {
    Ok((rf, ri)) if ri.zero < first_zero => { /* adopt */ }
    _ => {
        mc64_retry_not_adopted_now = true;
        Ok((factors, inertia))
    }
}
```

The `_` arm absorbs `Err(_)`. So when the first factorization returns
`inertia.zero > 0` and the issue-#65 retry runs, an interrupt observed *during
that retry* has two consequences, neither intended:

1. `factor` returns `Success` / `Singular` rather than `Interrupted`. That
   contradicts the contract this PR publishes — "the first observation of
   `true` aborts the factorization" — and the host's cancellation silently does
   not take effect for that call.

2. `self.mc64_retry_not_adopted` latches. That latch is keyed on the *pattern*
   and cleared only on a pattern change (`src/numeric/solver.rs:1038`). An
   interior-point host refactors a **fixed** pattern for the whole solve, so a
   single unrelated cancellation disables the MC64 rescue for every remaining
   iterate.

(2) is the serious half. The field's own doc-comment (`src/numeric/solver.rs:252-273`)
already flags this latch as interacting with the inertia hard rule:

> A pattern singular at iterate `k` (retry not adopted, latch set) that becomes
> MC64-rescuable at `k+1` on the *same pattern with different values* has its
> retry suppressed — reporting unrescued inertia where the pre-latch code
> recovered.

That doc reasons about the latch being armed by a *genuine* non-adoption, where
"MC64 didn't help" is real evidence. Here it is armed by a cancellation, where
no evidence was gathered at all — the retry never finished. The result is that
cancelling a factorization, which the PR correctly describes as leaving no
trace beyond the invalid factors, can instead cause wrong inertia for the
remainder of a solve. That is the one hard constraint in `CLAUDE.md`.

Fix is one arm, before the `_`:

```rust
match retry {
    Ok((rf, ri)) if ri.zero < first_zero => { /* adopt */ }
    // Issue #194: a cancellation is not evidence that MC64 does not
    // help -- the retry never got to finish. Propagate it, and leave
    // the per-pattern latch disarmed so the rescue is still available
    // on the next factor once the caller clears its flag.
    Err(FeralError::Interrupted) => Err(FeralError::Interrupted),
    _ => {
        mc64_retry_not_adopted_now = true;
        Ok((factors, inertia))
    }
}
```

Worth a test alongside the existing eight: arm the flag so that it is observed
during the retry, and assert both that `factor` reports `Interrupted` and that
`mc64_retry_attempt_count()` / the rescue still fire on the next `factor` after
the flag is cleared.

### Two smaller things

**The CHANGELOG's "Additive" heading undersells the break.** Neither
`FeralError` nor `FactorStatus` is `#[non_exhaustive]`, so every downstream
exhaustive `match` fails to compile. The bullet does say "exhaustive `match`es
over them need a new arm", which is accurate — but it sits under a heading that
says the opposite, and a reader skimming headings gets the wrong answer. This is
the same shape as the claim that had to be retracted on #191 ("existing callers
compile unchanged"), and it was a downstream build that caught that one. Suggest
labelling it **BREAKING for downstream matches** and keeping the prose as-is.

**The CHANGELOG will conflict.** This branch adds a new `## [Unreleased]`
header, but `main` now has one (from #191 and #193). Fold the entry into the
existing section on rebase rather than leaving two.

### Benchmark gate

Not a blocker for this PR to resolve — but for the record, the corpus is
available here and I have just re-run the Phase 2.8.1 partition interleaved
against `main` for #191 on the same host. Both exit partitions PASS on `main`
at dense 1.60-1.62 / 2.00 and sparse 1.58-1.64. So the session 2026-08-19-05
p90s this PR could not confirm are, as of now, confirmed on `main`; #195 still
needs its own run, but the baseline it would be measured against is known good.

jkitchin:
## Review

The design is right and the write-up is unusually honest — flagging the unmet
benchmark gate, and flagging that the first un-alternated A/B block was
container drift rather than reporting the 40% "regression", is exactly the
standard this repo asks for. Two poll granularities rather than one is the
correct call, and the issue-#8 justification for it is concrete.

I checked the mechanism itself and it holds up:

- Ordering escalation (`src/numeric/solver.rs:1558`) sits on the `Ok` path, so
  `Interrupted` cannot drive the recursive re-`factor`.
- `run_parallel_task`'s `task_res = Err(..); break` lands in the pre-existing
  `Err` arm (`src/numeric/factorize.rs:3919`), which sets `first_error` and
  skips the parent enqueue — and it is a `break`, not a `return`, so the
  `local_contribs` restore into the workspace still runs. That distinction
  matters and it was gotten right.
- Explicit arms instead of wildcards across the blast radius is the right
  trade, for the stated reason.

One defect, and it is worth fixing before merge.

### The MC64 rescue swallows `Interrupted`, and latches

`src/numeric/solver.rs:1359`:

```rust
match retry {
    Ok((rf, ri)) if ri.zero < first_zero => { /* adopt */ }
    _ => {
        mc64_retry_not_adopted_now = true;
        Ok((factors, inertia))
    }
}
```

The `_` arm absorbs `Err(_)`. So when the first factorization returns
`inertia.zero > 0` and the issue-#65 retry runs, an interrupt observed *during
that retry* has two consequences, neither intended:

1. `factor` returns `Success` / `Singular` rather than `Interrupted`. That
   contradicts the contract this PR publishes — "the first observation of
   `true` aborts the factorization" — and the host's cancellation silently does
   not take effect for that call.

2. `self.mc64_retry_not_adopted` latches. That latch is keyed on the *pattern*
   and cleared only on a pattern change (`src/numeric/solver.rs:1038`). An
   interior-point host refactors a **fixed** pattern for the whole solve, so a
   single unrelated cancellation disables the MC64 rescue for every remaining
   iterate.

(2) is the serious half. The field's own doc-comment (`src/numeric/solver.rs:252-273`)
already flags this latch as interacting with the inertia hard rule:

> A pattern singular at iterate `k` (retry not adopted, latch set) that becomes
> MC64-rescuable at `k+1` on the *same pattern with different values* has its
> retry suppressed — reporting unrescued inertia where the pre-latch code
> recovered.

That doc reasons about the latch being armed by a *genuine* non-adoption, where
"MC64 didn't help" is real evidence. Here it is armed by a cancellation, where
no evidence was gathered at all — the retry never finished. The result is that
cancelling a factorization, which the PR correctly describes as leaving no
trace beyond the invalid factors, can instead cause wrong inertia for the
remainder of a solve. That is the one hard constraint in `CLAUDE.md`.

Fix is one arm, before the `_`:

```rust
match retry {
    Ok((rf, ri)) if ri.zero < first_zero => { /* adopt */ }
    // Issue #194: a cancellation is not evidence that MC64 does not
    // help -- the retry never got to finish. Propagate it, and leave
    // the per-pattern latch disarmed so the rescue is still available
    // on the next factor once the caller clears its flag.
    Err(FeralError::Interrupted) => Err(FeralError::Interrupted),
    _ => {
        mc64_retry_not_adopted_now = true;
        Ok((factors, inertia))
    }
}
```

Worth a test alongside the existing eight: arm the flag so that it is observed
during the retry, and assert both that `factor` reports `Interrupted` and that
`mc64_retry_attempt_count()` / the rescue still fire on the next `factor` after
the flag is cleared.

### Two smaller things

**The CHANGELOG's "Additive" heading undersells the break.** Neither
`FeralError` nor `FactorStatus` is `#[non_exhaustive]`, so every downstream
exhaustive `match` fails to compile. The bullet does say "exhaustive `match`es
over them need a new arm", which is accurate — but it sits under a heading that
says the opposite, and a reader skimming headings gets the wrong answer. This is
the same shape as the claim that had to be retracted on #191 ("existing callers
compile unchanged"), and it was a downstream build that caught that one. Suggest
labelling it **BREAKING for downstream matches** and keeping the prose as-is.

**The CHANGELOG will conflict.** This branch adds a new `## [Unreleased]`
header, but `main` now has one (from #191 and #193). Fold the entry into the
existing section on rebase rather than leaving two.

### Benchmark gate

Not a blocker for this PR to resolve — but for the record, the corpus is
available here and I have just re-run the Phase 2.8.1 partition interleaved
against `main` for #191 on the same host. Both exit partitions PASS on `main`
at dense 1.60-1.62 / 2.00 and sparse 1.58-1.64. So the session 2026-08-19-05
p90s this PR could not confirm are, as of now, confirmed on `main`; #195 still
needs its own run, but the baseline it would be measured against is known good.

---

## Applied on this branch (7b42414)

All three of the above are fixed, and `origin/main` (now carrying #191 and
#193) is merged in at 942d130 — the CHANGELOG/decisions conflicts resolved by
keeping both sides under the single existing `## [Unreleased]`, and
`dev/context.md` regenerated rather than hand-merged.

The regression test is `tests/issue194_interrupt_during_mc64_retry.rs`. Hitting
the retry window needs care, so a note on how it does it: the PR's own tests
warm the symbolic cache with a first `factor()`, but that is unavailable here —
warming requires factoring the singular pattern, which arms the very latch under
test, and there is no public API to clear it. So the test calibrates the full
call time on a fresh solver, then sweeps the flag-set delay across it in 1%
steps with a spin-wait watchdog (`thread::sleep`'s ~1 ms granularity overshoots
a ~6 ms call).

The important detail is that the window detector is deliberately **independent
of the behaviour under test**: `mc64_retry_attempt_count() == 1` proves
factorization #1 returned `Ok(..)` (the gate keys on `Ok`, so the flag was set
after it completed), and `delay < call_elapsed` proves it was set before the
call returned. Together they pin the flag inside the retry without referencing
the status being asserted. An earlier version keyed the detector on
`status == Interrupted` — which is precisely what the bug swallows, so pre-fix
it could never report "landed" and would have misreported a live bug as "never
exercised". It ends on a hard vacuity guard rather than passing silently if no
attempt lands.

RED was confirmed verbatim before the fix went in:

```
landed inside the retry at frac=0.52 (delay 5.296698ms of call 6.352291ms) after 13 attempts; status Success
panicked: an interrupt observed during the MC64 retry must be reported as
Interrupted, not swallowed into the original factor's status; got Success
```

GREEN after, stable across 8 consecutive release runs plus a debug run. Full
`cargo test --all`: 0 failures.

One correction to my own framing above: I called this a race, and it is not
much of one. The vulnerable window is the entire duration of the retry — a
second full factorization of the same matrix — which is exactly the
long-factorization case issue #194 exists to interrupt.

jkitchin:
Review addressed. The MC64 finding was real — I verified it against the code before acting, and we converged: you pushed `7b42414` while I was preparing the same one-arm fix. Yours is on the branch and mine is dropped, because yours is better on two counts: the fixture is a ~6 ms call rather than my ~475 ms one, and it additionally covers the interrupt-during-factorization-#1 case that mine did not.

Pushed `f112b93` on top of yours (fast-forward, nothing rewritten) with what was left.

## The test trap is worth a note, because we both hit it

My first version of that regression test **passed against the un-fixed code**. The detector for "did the watchdog land inside the rescue?" was `status == Interrupted && retry_attempt_count == 1` — and `status == Interrupted` is precisely what the bug suppresses. On un-fixed code every trial read as a miss, the sweep fell through to its skip-and-pass guard, and it went green against the defect it was written for:

```
  65% (359.9ms): missed the window — status Success, retry_attempts 1
  75% (415.3ms): missed the window — status Success, retry_attempts 1
  SKIP: no watchdog fraction landed inside the MC64 rescue
```

Every one of those "missed" lines is the bug firing, reported as a miss. Your commit message records the same trap and the same repair, independently — which is what makes it worth writing down as a rule rather than a one-off: **a skip-and-pass guard must never be reachable through the failure mode it guards.** This tree leans on skip-and-pass widely for fixture- and timing-gated tests, so it may be worth an audit. Recorded in the journal and the checkpoint.

## Also in `f112b93`

**The session records now cover the review round.** The journal and checkpoint stopped at the benchmark artifact and had nothing on the MC64 defect, the test trap, the merge, or the CI resolution — CLAUDE.md wants those in the journal in real time, so that was a genuine gap.

**`tests/issue194_factor_interrupt.rs` cost 46.6 s in the debug build CI actually runs** (`cargo test`, ci.yml:37) — nearly all of it the two mid-flight tests at `grid_laplacian(350)`, n = 122500, four factorizations each. That was mine and I had not noticed it: I calibrated those tests in `--release`, where debug is ~30x slower. Dropped to n = 90000: **46.6 s → 21.8 s** debug, 0.85 s release, all eight tests still exercised and **none skipped in either mode** (release warm baselines 142 ms parallel / 233 ms sequential, both above the 100 ms anti-flake guard).

## Your two smaller points

**The "Additive" heading** — you fixed it in `7b42414` ("BREAKING for exhaustive matches; additive at runtime"). I had made the same relabel during conflict resolution; yours survived the merge and reads better. Nothing further from me.

**The CHANGELOG conflict** — resolved in `942d130`/my merge: one `[Unreleased]` section with #192 first, then #194. `dev/decisions.md` kept both sides' entries chronologically (append-only, nothing edited), and `dev/context.md` was regenerated with `dev/assemble-context.sh` rather than hand-resolved.

## CI

The `check` blocker resolved itself upstream: `9c5dfac` on `main` is exactly the clippy 1.98 drift I had stood down on. The branch carries `main` (`git merge-base --is-ancestor 205cbbb HEAD` confirms it is current).

I have since installed the CI toolchain locally, so that gate is now reproducible before pushing instead of being discovered in CI — the container default is 1.94.1, which does not have the lints that were failing. Pre-push on `f112b93`, with 1.98.0: `fmt --check` clean, `clippy --all-targets -- -D warnings` clean for both `feral` and `feral-diagnostics`, full suite **103 binaries / 979 passed / 0 failed**.

## Benchmark gate

Thank you for the `main` baseline — that closes the half I could not. The corpus still is not in this container, so this branch's own Phase 2.8.1 run remains outstanding and the PR does not claim otherwise; it is item 1 in the checkpoint's next-session list.

---

